PR #104, opened 2017-08-22
Fix root CMake should make tensorflow2ncnn

Corea:


nihui:
tensorflow2ncnn is experimental and still under construction. It will be included in the default build once I think it is ready.

Corea:
Okay, I understood. 
Besides, I would like to change `build.sh` and `package.sh` with `/usr/bin/env`. How do you think?

PR #130, opened 2017-09-07
undate softmax.cpp

vicwer:
make O(n) = 1，save the memory qwq


nihui:
axis support implemented in softmaxv2

PR #143, opened 2017-09-20
fix tailpad

echofeng:
fix tail pad error ,when input size is smaller than kernel size


ArmstrongYang:
@nihui I also find this bug in the latest version ncnn, when do you merge the pull requests?

PR #144, opened 2017-09-20
bug fix for conv1x1s2 on ARMv8 neon

Corea:
I'm not sure with assembly code below, which is for ARMv7, so please do double check this function.


nihui:
there is no error in armv7 assembly code for this scope

PR #145, opened 2017-09-20
fix the apply var bug in mvn layer

lzx1413:
the bottom data do not subtract the mean, so we should use the top data .


nihui:
so line 185 should be changed as well ... emmmm...
useless lines should be removed if not used
```
const float* ptr = bottom_blob.channel(q);
```

lzx1413:
aha,you are right

PR #155, opened 2017-10-08
add interp layer support nearest and bilinear

lzx1413:


nihui:
assign raw data pointer directly is wrong
check size equality before creating top_blob
and use top_blob = bottom_blob for the same size
```
       if(oh == h && ow == w) {
           top_blob.data = bottom_blob.data;
           return 0;
       }
```

please follow the coding style of the existing layer implementation, such as no space idents within namespace scope.

lzx1413:
Ok I'll fix it soon

lzx1413:
should I move the neon part code into arm folder?

PR #156, opened 2017-10-09
add mobilenet ssd example

lzx1413:


examples/ssd/ssdmobilenet.cpp
```diff
@@ -0,0 +1,111 @@
...
+    auto input_size = 300;
+    ncnn::Mat in = ncnn::Mat::from_pixels_resize(raw_img.data, ncnn::Mat::PIXEL_BGR, raw_img.cols, raw_img.rows, input_size, input_size);
+
+    const float mean_vals[3] = {127.5f, 127.5f, 127.5f};
+    const float norm_vals[3] = {1.0/127.5,1.0/127.5,1.0/127.5};
+    in.substract_mean_normalize(mean_vals, norm_vals);
+
+    ncnn::Mat out;
+
+    for(int i=0;i<1;i++)
```
nihui: this loop is useless

examples/ssd/ssdmobilenet.cpp
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    printf("%d %d %d\n", out.w, out.h, out.c);
+    std::vector<Object> objects;
+    for (int ih=0;ih<out.h;ih++)
+    {
+        Object object;
+        for(int iw=0;iw<out.w;iw++)
+        {
+            auto value = out.data[iw+ih*out.w];
```
nihui: out.w is always six, so it's fine to write const float* values = out.row(ih); and use values[0] values[1] .... instead of the following six conditions
do not use auto keyword

examples/ssd/ssdmobilenet.cpp
```diff
@@ -0,0 +1,111 @@
...
+                object.rec.width = value*img_w-object.rec.x;
+            if(iw == 5)
+                object.rec.height = value*img_h-object.rec.y;
+            printf("%f ", out.data[iw+ih*out.w]);
+        }
+        objects.push_back(object);
+        printf("\n");
+
+    }
+    for(auto object : objects)
```
nihui: always prefer indexed for-loop over c++11 range-based for

examples/ssd/ssdmobilenet.cpp
```diff
@@ -0,0 +1,98 @@
...
+    ncnn::Net mobilenet;
+    /*
+     * model is  converted from https://github.com/chuanqi305/MobileNet-SSD
+     * and can be downloaded from https://drive.google.com/open?id=0ByaKLD9QaPtucWk0Y0dha1VVY0U
+     */
+    int img_h = raw_img.size().height;
+    int img_w = raw_img.size().width;
+    mobilenet.load_param("mobilenet_ssd_voc_ncnn.param");
+    mobilenet.load_model("mobilenet_ssd_voc_ncnn.bin");
+    auto input_size = 300;
```
nihui: no c++11 auto here

examples/ssd/ssdmobilenet.cpp
```diff
@@ -0,0 +1,98 @@
...
+        object.rec.width = values[4] * img_w - object.rec.x;
+        object.rec.height = values[5] * img_h - object.rec.y;
+        objects.push_back(object);
+    }
+
+    for(int i = 0;i<objects.size();++i)
+    {
+        Object object = objects.at(i);
+        if(object.prob > show_threshold)
+            cv::rectangle(raw_img,object.rec,cv::Scalar(255,0,0));
```
nihui: It will be nice to have class-id and score text drawn besides rectangle only

src/layer/interp.cpp
```diff
@@ -25,13 +25,13 @@ Interp::Interp()
 
 Interp::~Interp()
 {
-}
```
nihui: interp.cpp file conflicts, please fix it.

PR #160, opened 2017-10-12
Add epsilon on variance of batch normalization layer.

Corea:
caffe (https://github.com/BVLC/caffe/blob/master/src/caffe/layers/batch_norm_layer.cpp#L150) and Tensorflow (https://github.com/tensorflow/tensorflow/blob/754048a0453a04a761e112ae5d99c149eb9910dd/tensorflow/core/kernels/batch_norm_op.h#L117) add epsilon value to variance.


Corea:
Sorry, I didn't check caffe2ncnn code. Then I will add some code to process `FusedBatchNorm` on tensorflow2ncnn, and also put similar code on them. thanks!

PR #200, opened 2017-11-22
fix unaryop bug

wind19870521:
- incorrect memory access when matrix is multi-channel

Signed-off-by: wangshunli <shunli0521.hi@163.com>


nihui:
int size = w * h * channels; 改为 int size = a.total(); 会更好，避免循环 channel

PR #218, opened 2017-12-18
fix mobilenet: add depthwise, fix batch norm

vsooda:
mobilenet is work now

### DIFF
--- a/tools/mxnet/mxnet2ncnn.cpp
+++ b/tools/mxnet/mxnet2ncnn.cpp
@@ -548,6 +548,10 @@ static bool read_mxnet_param(const char* parampath, std::vector<MXNetParam>& par
         {
             p.name = std::string(p.name.c_str() + 4);
         }
+        if (memcmp(p.name.c_str(), "aux:", 4) == 0)
+        {
+            p.name = std::string(p.name.c_str() + 4);
+        }
 
 //         fprintf(stderr, "%s read\n", p.name.c_str());
     }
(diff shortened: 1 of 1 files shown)


tools/mxnet/mxnet2ncnn.cpp
```diff
@@ -838,6 +847,9 @@ int main(int argc, char** argv)
         else if (n.op == "BatchNorm")
         {
             float eps = n.attr("eps");
+            if (fabs(eps) < 1e-6) {
```
nihui: why eps should be changed here

vsooda: the default eps in mxnet is 1e-3. in some version, it is not writed in symbol.json

like this: 

```
{
      "op": "null",
      "name": "batchnorm18_gamma",
      "inputs": []
    },
    {
      "op": "null",
      "name": "batchnorm18_beta",
      "inputs": []
    },
    {
      "op": "null",
      "name": "batchnorm18_moving_mean",
      "attr": {"__init__": "[\"zero\", {}]"},
      "inputs": []
    },
    {
      "op": "null",
      "name": "batchnorm18_moving_var",
      "attr": {"__init__": "[\"one\", {}]"},
      "inputs": []
    },
    {
      "op": "BatchNorm",
      "name": "batchnorm18",
      "inputs": [[165, 0, 0], [166, 0, 0], [167, 0, 0], [168, 0, 1], [169, 0, 1]]
    },
```

nihui: ok, but it could be checked using
```n.has_attr("eps")```

vsooda: updated

PR #222, opened 2017-12-24
add priorbox cache to accelerate the ssd detection

lzx1413:


nihui:
layer implementation should be state-less and cares the inference process only by design
intermediate result cache could be handled at the caller side
the following snippet works in both light mode and non-light mode
```
// loaded net
ncnn::Net net;

// you want to cache the intermediate blob "bbox1"
ncnn::Mat bbox1cache;

// dummy loop
for (int i=0; i<100; i++)
{
    // input
    ncnn::Net in;
    ncnn::Net out;

    ncnn::Extractor ex = net.create_extractor();
    ex.set_light_mode(true);
    ex.input("data", in);

    // reuse cache or generate cache
    if (bbox1cache.empty())
    {
        ex.extract("bbox1", bbox1cache);
    }
    else
    {
        ex.input("bbox1", bbox1cache);
    }

    ex.extract("output", out);
}
```

PR #253, opened 2018-01-31
fix conv3x3s1 winograd_neon3

CamelliaJune:


nihui:
hello
already fixed in https://github.com/Tencent/ncnn/commit/c6506d6ecd7b9dbd7965e2b79a4fab18504b6901

thanks anyway ;)

PR #306, opened 2018-03-26
bugfix

chenqy4933:
binding the neon variables to fix register


nihui:
please keep the ident style

PR #311, opened 2018-03-27
Add the epsilon parameter to the BatchNorm layer. (fix #303)

### LINKED ISSUE #303: BatchNorm Layer: it seems to be safer to add a epsilon to the standard deviation
In the following codes cited from the file [src/layer/batchnorm](https://github.com/Tencent/ncnn/blob/6ea09ebf2c4cc05460a99dea53a6260f5843c19a/src/layer/batchnorm.cpp#L62), if the `sqrt_var` is zero, there will be a zero-division error. So I advice to add a epsilon to the standard deviation or add some protection codes.

        float sqrt_var = sqrt(var_data[i]);
        a_data[i] = bias_data[i] - slope_data[i] * mean_data[i] / sqrt_var;
        b_data[i] = slope_data[i] / sqrt_var;

(虽然可能性不大, 但加上去毕竟更安全些)

cook:
Liking the layers MVN and InstanceNorm, add the epsilon to the BatchNorm layer as a parameter to avoid the possible zero-division error.

* Making the default value of the `eps` to be `0.f` is to compatible with the old version.

### DIFF
--- a/src/layer/batchnorm.cpp
+++ b/src/layer/batchnorm.cpp
@@ -60,7 +60,7 @@ int BatchNorm::load_model(const ModelBin& mb)
 
     for (int i=0; i<channels; i++)
     {
-        float sqrt_var = sqrt(var_data[i]) + eps;
+        float sqrt_var = sqrt(var_data[i] + eps);
         a_data[i] = bias_data[i] - slope_data[i] * mean_data[i] / sqrt_var;
         b_data[i] = slope_data[i] / sqrt_var;
     }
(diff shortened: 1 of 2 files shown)


src/layer/batchnorm.cpp
```diff
@@ -59,7 +60,7 @@ int BatchNorm::load_model(const ModelBin& mb)
 
     for (int i=0; i<channels; i++)
     {
-        float sqrt_var = sqrt(var_data[i]);
+        float sqrt_var = sqrt(var_data[i]) + eps;
```
nihui: eps should be added inside sqrt()
ref https://arxiv.org/abs/1502.03167

PR #317, opened 2018-03-29
fix mem.h start address

Howave:


nihui:
hello, MSVC has its __declspec(align(4))
please let it be more portable

PR #320, opened 2018-03-30
add image-level feature support

Lamply:
Resize Cx1x1 feature map to CxHxW, use in DeepLabv3 etc.

### DIFF
--- a/src/layer/interp.cpp
+++ b/src/layer/interp.cpp
@@ -42,6 +42,12 @@ int Interp::forward(const Mat &bottom_blob, Mat &top_blob) const
     int c = bottom_blob.c;
     int oh = output_height;
     int ow = output_width;
+    if (bottom_blob.dims == 1)
+    {
+        h = 1;
+        w = 1;
+        c = bottom_blob.w;
+    }
     if (ow == 0 || ow == 0)
     {
         oh = h * height_scale;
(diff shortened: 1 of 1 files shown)


src/layer/interp.cpp
```diff
@@ -56,26 +62,42 @@ int Interp::forward(const Mat &bottom_blob, Mat &top_blob) const
     if (top_blob.empty())
         return -100;
 
-    if (resize_type == 1)//nearest
+    if (resize_type == 1 || bottom_blob.dims == 1)//nearest
```
nihui: If bottom_blob is a vector, the resize_type is ignored. move this condition out of resize_type path.

```
if (bottom_blob.dims == 1)
{
    // handle this special case, ignore resize_type

    return 0;
}

// all the following code shall not be changed
if (resize_type == 1)//nearest
{
}

```

src/layer/interp.cpp
```diff
@@ -56,26 +62,42 @@ int Interp::forward(const Mat &bottom_blob, Mat &top_blob) const
...
-        #pragma omp parallel for
-        for (int q = 0; q < c; ++q)
+        if(bottom_blob.dims == 1)
+        {
+            #pragma omp parallel for
+            for (int q = 0; q < c; ++q)
+            {
+                const float *ptr = ((float*)bottom_blob.data+q);
+                float *output_ptr = top_blob.channel(q);
+                for (int i = 0; i < top_blob.cstep; ++i)
```
nihui: This is what the Mat::fill() does

src/layer/interp.cpp
```diff
@@ -56,26 +62,38 @@ int Interp::forward(const Mat &bottom_blob, Mat &top_blob) const
     if (top_blob.empty())
         return -100;
 
+	if (bottom_blob.dims == 1)
+	{
+		#pragma omp parallel for
```
nihui: always use space, not tab

src/layer/interp.cpp
```diff
@@ -56,26 +62,38 @@ int Interp::forward(const Mat &bottom_blob, Mat &top_blob) const
...
+		return 0;
+	}
+	
     if (resize_type == 1)//nearest
     {
-        #pragma omp parallel for
-        for (int q = 0; q < c; ++q)
-        {
-            const float *ptr = bottom_blob.channel(q);
-            float *output_ptr = top_blob.channel(q);
```
nihui: keep the ident space of this block

src/layer/interp.cpp
Lamply: Oh, thanks

PR #328, opened 2018-04-01
Fix Mat::fill when parameter is int or double

clcarwin:


nihui:
The fix is wrong.
Template version of Mat::fill is intended to fill int/uchar/double.
Pass float type argument explicitly if you want to set with float type.
```
Mat a;
Mat b;
a.fill(1.f);
b.fill((float)3.1415);
```

clcarwin:
If parameter is double, the buffer will be overflowed.

nihui:
create a Mat with element of double type, be aware of the elemsize argument
```
// initialize 10x10x4 Mat with double element type, and fill with 3.1415, double type
int w = 10;
int h = 10;
int c = 4;
Mat m(w, h, c, sizeof(double));
m.fill(3.1415);

// pointer access
const double* readonly_ptr = m;
double* readwrite_ptr = m;
```

clcarwin:
Thanks, I have understood the design. In my opinion it is better to check sizeof(T) and elemsize. 
```if(sizeof(T) > elemsize) error()```

PR #337, opened 2018-04-08
update benchncnn

BUG1989:
update the benchmark of qcom820


nihui:
https://github.com/Tencent/ncnn/commit/2ecad8ab32f1e2f62c70ccc0f138ce511875a35e

PR #541, opened 2018-08-23
Add loading model param directly from memory.

lygstate:
# Conflicts:
#	src/net.cpp
#	src/paramdict.cpp

# Conflicts:
#	src/net.cpp

### DIFF
--- a/src/net.cpp
+++ b/src/net.cpp
@@ -94,12 +94,10 @@ int Net::register_custom_layer(int index, layer_creator_func creator)
     return 0;
 }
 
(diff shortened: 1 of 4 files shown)


nihui:
only binary format memory loading is supported by design
If you want to embed your graph and model weight into code, stick to binary format
otherwise you will get larger binary, and all plain string literals are visible inside your binary ;)

| load from | XYZ.param | XYZ.param.bin | XYZ.bin |
|---|---|---|---|
|file|y|y|y|
|memory|n|y|y|

PR #648, opened 2018-11-09
optimize float2int8 conversion

biaoxiaoduan:
这种写法应该会比之前的写法快一些，有空的话可以试试~

### DIFF
--- a/src/layer/quantize.cpp
+++ b/src/layer/quantize.cpp
@@ -35,10 +35,14 @@ int Quantize::load_param(const ParamDict& pd)
 
 static inline signed char float2int8(float v)
 {
-    int int32 = round(v);
-    if (int32 > 127) return 127;
-    if (int32 < -128) return -128;
-    return (signed char)int32;
+    // round
+    int int32 = v+128.5f;
+    
+    // clip: > 255 or < 0
+    if (int32 & (~0xFF)) {
+        int32 = (-int32) >> 31;
+    }
+    return (signed char)(int32-128);
 }
 
 int Quantize::forward(const Mat& bottom_blob, Mat& top_blob, const Option& opt) const


tencent-adm:
[CLA assistant check](https://cla.opensource.tencent.com/Tencent/ncnn?pullRequest=648) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.opensource.tencent.com/Tencent/ncnn?pullRequest=648) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.opensource.tencent.com/check/Tencent/ncnn?pullRequest=648) it.</sub>

nihui:
is there any benchmark result ?

biaoxiaoduan:
```c++
#include <stdio.h>
#include <stdlib.h>
#include <sys/time.h>
#include <math.h>

static inline signed char float2int8_old(float v){
    int int32 = round(v);
    if (int32 > 127) return 127;
    if (int32 < -128) return -128;
    return (signed char)int32;
}

static inline signed char float2int8_new(float v)
{
    // round
    int int32 = v+128.5f;
    // clip: > 255 or < 0
    if (int32 & (~0xFF)) {
        int32 = (-int32) >> 31;
    }
    return (signed char)(int32-128);
}

int main(int argc, char **argv) {
    
    const int num = 10000 * 1000;
    float *src = new float[num];
    signed char *dst1 = new signed char[num];
    signed char *dst2 = new signed char[num];

    // generate random float array value between -300~300
    for (int i = 0; i < num; i++) {
        int rd = rand() % 6000;
        rd = rd - 3000;
        src[i] = rd/10.0;
    }

    struct timeval time1, time2, time3;
    gettimeofday(&time1, NULL);
    // test old
    for (int i = 0; i < num; i++) {
        dst1[i] = float2int8_old(src[i]);
    }
    gettimeofday(&time2, NULL);

    // test new
    for (int i = 0; i < num; i++) {
        dst2[i] = float2int8_new(src[i]);
    }
    gettimeofday(&time3, NULL);

    long intervalOld = (time2.tv_sec-time1.tv_sec) * 1000000 + time2.tv_usec-time1.tv_usec;
    long intervalNew = (time3.tv_sec-time2.tv_sec) * 1000000 + time3.tv_usec-time2.tv_usec;
    printf("old algorithm cost %ld us\n", intervalOld);
    printf("new algorithm cost %ld us\n", intervalNew);
    
    bool success = true;
    for (int i = 0; i < num; i++) {
        if (dst1[i] != dst2[i]) {
            printf("%d: src=%f, dst1=%d, dst2=%d\n", i, src[i], dst1[i], dst2[i]);
            success = false;
            break;
        }
    }
    if (success) {
        printf("result is correct\n");
    } else {
        printf("result is incorrect\n");
    }
    
    delete [] src;
    delete [] dst1;
    delete [] dst2;
    src = NULL;
    dst1 = NULL;
    dst2 = NULL;
 }

```

I used the code above to test performance.
The build script is 
```
g++ -o main main.cpp -O3
```
On my Lenovo S5 (Intel i7 7700HQ), result is as below
```
old algorithm cost 68746 us
new algorithm cost 7431 us
5: src=-66.500000, dst1=-67, dst2=-66
result is incorrect
```

I think there is a major acceleration.
BTW, I noticed that the result is slightly different. -66.5 become -66 on your end, and it's -67 on my end. I am not sure which one is right.

nihui:
@BUG1989 Could you please review this PR?

BUG1989:
@biaoxiaoduan 
emmm......-66.5 become -66 on your end, and it's -67 on ncnn's end.
on the implement of arm platform,we use "FCVTAS" to "Floating-point convert to signed integer, rounding to nearest with ties to away",and it's -67 on ncnn-arm's end.

nihui:
the naive C implementation is not intended for speed optimized but for reference only

closed

feel free to create another new pull request, if you move these optimization into x86 folder and fixed the rounding problem ;)

PR #682, opened 2018-11-30
fix the model load error after converting model by ncnn2mem  

neos2008:
if implement silence layer outside, the model load error without any warring. And make the silence layer integrate with ncnn library, it can fix this problem.


nihui:
silence layer is commonly used in training.
the common practice is to remove the silence layer in your deploy.prototxt, and then converting your model to ncnn.

It would be better implementing a common "Noop" layer , and then conerting all the training operators to Noop

PR #684, opened 2018-12-03
fix: avoid rebuilding after cmake

ayounes-nviso:
Using file() was triggering a recompilation each time cmake was called. This
commit avoids void using string() configure_file() instead.


ayounes-nviso:
Thanks for accepting it :-)

PR #693, opened 2018-12-06
The compiler needs to be using C++11

toniz:
https://stackoverflow.com/questions/52728417/osx-caffe-compilation-fails-with-expected-expression-error


nihui:
only model converters needs c++11 for newer protobuf version
fixed in https://github.com/Tencent/ncnn/commit/0ca92de9359e37db20c86a9f93d4858bfb5afd93

Issue #697, opened 2018-12-10
现在还不支持lstm吗？网络中包含lstm报错


z13974509906:
我的caffe prototxt中有LSTM， 使用caffe2ncnn可以把模型转换成ncnn的param和bin文件，但是导入param和bin文件的时候报错

>layer LSTM not exists or registered

是ncnn现在还不支持LSTM吗，但是我看src中是有lstm.cpp和lstm.h的

求解决办法！嘤嘤嘤

ppl21847:
@

> 自定义

麻烦问一下 怎样自定义呀？

beniz:
Hi, it seems LSTM has remained set to OFF in https://github.com/Tencent/ncnn/blob/master/src/CMakeLists.txt#L162 and for me this is what generates the `layer LSTM not exists or registered` error.

Any particular reason for not having LSTM to ON by default ?

beniz:
After looking at it carefully, LSTM layer does not work (for me) at a deeper level. Converting from a Caffe model, there are typically differences in the acquisition of input dimensions.

Here, the input height `h` is used as the timestep https://github.com/Tencent/ncnn/blob/master/src/layer/lstm.cpp#L68 whereas Caffe checks on the first dimension of the blob (dim 0).

My current testbed is a CRNN I can actually share, and the output of convolutions does not fit this LSTM layer whereas it works with Caffe. I've been playing with the `lstm.cpp` code, and fixing the dimension acquisition above does not seem to be enough. It's possible there are additional bugs.

z13974509906:
> After looking at it carefully, LSTM layer does not work (for me) at a deeper level. Converting from a Caffe model, there are typically differences in the acquisition of input dimensions.
> 
> Here, the input height `h` is used as the timestep https://github.com/Tencent/ncnn/blob/master/src/layer/lstm.cpp#L68 whereas Caffe checks on the first dimension of the blob (dim 0).
> 
> My current testbed is a CRNN I can actually share, and the output of convolutions does not fit this LSTM layer whereas it works with Caffe. I've been playing with the `lstm.cpp` code, and fixing the dimension acquisition above does not seem to be enough. It's possible there are additional bugs.
> 
> I guess it was OFF for a good reason :) I'll try to fix it.

I have test the lstm. For me, it can produce correct result.

beniz:
Thanks, see #782 for an init fix.

PR #782, opened 2019-02-11
fixed cell initialization in LSTM layer

beniz:
This fix allows LSTM to work correctly. 

The layer now works properly, and could be activated in https://github.com/Tencent/ncnn/blob/master/src/CMakeLists.txt#L162

This is a fix by @fantes.

See #697


tools/caffe/caffe.proto
```diff
@@ -359,15 +525,17 @@ message LayerParameter {
   // engine parameter for selecting the implementation.
   // The default for the engine is set by the ENGINE switch at compile-time.
   optional AccuracyParameter accuracy_param = 102;
+  optional AnnotatedDataParameter annotated_data_param = 200;
   optional ArgMaxParameter argmax_param = 103;
   optional BatchNormParameter batch_norm_param = 139;
   optional BiasParameter bias_param = 141;
-  optional BNParameter bn_param = 45;
```
nihui: do not drop BN layer for backward compability

tools/caffe/caffe.proto
```diff
@@ -594,17 +970,6 @@ message BiasParameter {
   optional FillerParameter filler = 3;
 }
 
-// Message that stores parameters used by BN (Batch Normalization)  layer
-message BNParameter {
```
nihui: do not drop BN layer for backward compability

tools/caffe/caffe2ncnn.cpp
```diff
@@ -321,18 +321,23 @@ static bool read_proto_from_binary(const char* filepath, google::protobuf::Messa
 
 int main(int argc, char** argv)
 {
-    if (!(argc == 3 || argc == 5 || argc == 6 || argc == 7))
+    int ocr = 0;
+
+    if (!(argc == 4 || argc == 6 || argc == 7 || argc == 8))
     {
-        fprintf(stderr, "Usage: %s [caffeproto] [caffemodel] [ncnnproto] [ncnnbin] [quantizelevel] [int8scaletable]\n", argv[0]);
+        fprintf(stderr, "Usage: %s ocr|noocr [caffeproto] [caffemodel] [ncnnproto] [ncnnbin] [quantizelevel] [int8scaletable]\n", argv[0]);
```
nihui: new optional parameter should always be placed as the last one for keeping the old usage works

src/layer/continuationindicator.cpp
```diff
@@ -0,0 +1,51 @@
...
+//
+// https://opensource.org/licenses/BSD-3-Clause
+//
+// Unless required by applicable law or agreed to in writing, software distributed
+// under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+// CONDITIONS OF ANY KIND, either express or implied. See the License for the
+// specific language governing permissions and limitations under the License.
+
+#include "continuationindicator.h"
+#include <iostream>
```
nihui: do not include the iostream header, no usecase

beniz:
@nihui apologies, we modified our local branch afterwards, thus affecting this PR. The fix is actually a one liner. Let's close it then we'll resubmit.

beniz:
See #784

PR #805, opened 2019-03-01
Add memorization of internal lstm state between different forward() calls

fantes:
This allows to do several forward() calls for same timeserie in order to have results on the fly. 
For instance, this allows to give input of time t0, call extract, get output for time t0 then give input for time t1, call extract, get output for time t1 and so on.   
Previously, internal state was reset at every forward() call.

### DIFF
--- a/src/layer/lstm.cpp
+++ b/src/layer/lstm.cpp
@@ -68,13 +68,14 @@ int LSTM::forward(const std::vector<Mat>& bottom_blobs, std::vector<Mat>& top_bl
     int size = input_blob.w;
 
     // initial hidden state
-    Mat hidden(num_output, 4u, opt.workspace_allocator);
+    // created only if necessary by Mat::create
+    hidden.create(num_output, 4u, opt.workspace_allocator);
     if (hidden.empty())
         return -100;
-    hidden.fill(0.f);
 
     // internal cell state
-    Mat cell(num_output, 4u, opt.workspace_allocator);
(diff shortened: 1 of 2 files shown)


tencent-adm:
[CLA assistant check](https://cla.opensource.tencent.com/Tencent/ncnn?pullRequest=805) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.opensource.tencent.com/Tencent/ncnn?pullRequest=805) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.opensource.tencent.com/check/Tencent/ncnn?pullRequest=805) it.</sub>

nihui:
It is wrong to put mutable data as class member because of potential data race in multi-threading forward.

In case you want to initialize hidden/cell with your own data, set/get them from bottom_blobs and top_blobs.

but I think this lstm layer unroll time series internally, doesn't it

because ncnn Mat support at-most 3-dimension data, it is possible to process each time tick from caller  (just like how ncnn handles multiple object proposals in faster-rcnn)
```
Mat hidden;
Mat cell;

for (int t=0; t<ticks; t++)
{
    ex.input("hidden", hidden);
    ex.input("cell", cell);

    ex.extract("hidden-out", hidden);
    ex.extract("cell-out", cell);
}

```

PR #841, opened 2019-03-19
add: cmake option to install NCNN SDK

ayounes-nviso:
A project using src/CMakeLists.txt directly does not want
to install NCNN library and headers. This new option makes it
optional (default to true).


ayounes-nviso:
Thanks Nihui for merging this patch, deeply appreciated!

Abdel

PR #844, opened 2019-03-20
 Apply splitcnn function to graph input

gemfield:
Some networks/models may split input to several inputs, in this situation, node_reference.size() will not match with the split line in ncnn.param file (generated by tools/onnx/onnx2ncnn).


nihui:
thanks a lot !

PR #848, opened 2019-03-21
Fix crash issue during load_model

gemfield:
Sometimes, convert tool may introduce inconsistent to ncnn.param file . For example, lay_count is 394 while only 393 layer followed. In this situation, calling load_model API will crash the runtime program.


nihui:
could you fix this issue in all load_model() functions,  load from memory

gemfield:
@nihui sure.

PR #948, opened 2019-04-25
Update onnx2ncnn.cpp

ShuangLiu1992:
num_filter for ConvTranspose is wrong (line 1075), should be the same sa Conv, which reads the number of filters from dim(0) (line 998)
int num_filter = W.dims(0);

### DIFF
--- a/tools/onnx/onnx2ncnn.cpp
+++ b/tools/onnx/onnx2ncnn.cpp
@@ -1072,7 +1072,6 @@ int main(int argc, char** argv)
         {
             const onnx::TensorProto& W = weights[node.input(1)];
 
-            int num_filter = W.dims(0);
             int has_bias = node.input_size() == 3 ? 1 : 0;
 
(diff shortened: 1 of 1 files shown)


ShuangLiu1992:
this fixes the number of filters but the number of output is not fixed. I guess this is a todo as indicated in the code: 
std::vector<int> output_shape = get_node_attr_ai(node, "output_shape");//TODO

I suppose the shape of the output tensor could be computed from consumers of this tensor by looking at the shape of their input tensor in weights?

ShuangLiu1992:
apparently I was wrong, it should be 
"The number of channels in the output should be equal to W.shape[1] * group (assuming zero based indices of the shape array)"
according to the Output section in https://github.com/onnx/onnx/blob/master/docs/Operators.md#ConvTranspose and https://github.com/Tencent/ncnn/pull/617
so maybe changing it to W.dims(1) * group should suffice?

To clarify we are testing with grouped Deconvolution which caused the conversion tool to crash

PR #1019, opened 2019-05-30
optimization the x86 convolution layer with avx2

BUG1989:
**optimization the x86 convolution layer**

- using a unprofessional sgemm to optimize the convolution layer
- the device should be supported avx2 instruction

**benchmark with single-thread**
image

### DIFF
--- a/src/layer/x86/convolution_7x7.h
+++ b/src/layer/x86/convolution_7x7.h
@@ -20,7 +20,7 @@ static void conv7x7s1_sse(const Mat &bottom_blob, Mat &top_blob, const Mat &_ker
     int stride_w = 1;
     int stride_h = 1;
 
-    conv_im2col_sgemm_sse(bottom_blob, top_blob, _bias, _kernel, kernel_w, kernel_h, stride_w, stride_h, opt);
+    conv_im2col_sgemm_sse(bottom_blob, top_blob, _kernel, _bias, kernel_w, kernel_h, stride_w, stride_h, opt);
 }
 
 static void conv7x7s2_sse(const Mat &bottom_blob, Mat &top_blob, const Mat &_kernel, const Mat& _bias, const Option& opt)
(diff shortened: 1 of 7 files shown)


src/layer/x86/convolution_x86.cpp
```diff
@@ -85,7 +86,7 @@ int Convolution_x86::create_pipeline(const Option& opt)
         int num_input = weight_data_size / 9 / num_output;
         // winograd is slow on small channel count
         if(num_input >= 16 && num_output >= 16)
-            use_winograd3x3 = true;
+            use_winograd3x3 = false;
```
nihui: intentional change ?

src/layer/x86/convolution_7x7.h
```diff
@@ -0,0 +1,35 @@
...
+
+static void conv7x7s1_sse(const Mat &bottom_blob, Mat &top_blob, const Mat &_kernel, const Mat& _bias, const Option& opt)
+{
+    int kernel_w = 7;
+    int kernel_h = 7;
+
+    int stride_w = 1;
+    int stride_h = 1;
+
+    conv_im2col_sgemm_sse(bottom_blob, top_blob, _bias, _kernel, kernel_w, kernel_h, stride_w, stride_h, opt);
```
nihui: _bias and _kernel interchange ooops ?

src/layer/x86/convolution_sgemm.h
```diff
@@ -0,0 +1,1028 @@
...
+// in compliance with the License. You may obtain a copy of the License at
+//
+// https://opensource.org/licenses/BSD-3-Clause
+//
+// Unless required by applicable law or agreed to in writing, software distributed
+// under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+// CONDITIONS OF ANY KIND, either express or implied. See the License for the
+// specific language governing permissions and limitations under the License.
+
+#include <immintrin.h>
```
nihui: this header file may be missing on non-x86 platform
```
#include "platform.h"
#if NCNN_AVX2
#include <immintrin.h>
#endif
```

src/layer/x86/convolution_x86.cpp
BUG1989: Yes!
Because the x86 winograd TODO optimizate.

PR #1045, opened 2019-06-12
fix neon intrinsic compile warnings

Howave:
the warning is as below:
ncnn/src/mat_pixel.cpp:1139:32: warning: implicit conversion from 'int' to 'int8_t' (aka 'signed char') changes
      value from 128 to -128 [-Wconstant-conversion]
    int8x8_t _v128 = vdup_n_s8(128);
                     ~~~~~~~~~ ^~~


nihui:

PR #1105, opened 2019-07-13
Tanh Support and fix#1103

wincle:
Tanh is an important op for image generation.
Fix problem #1103 . Which cause a problem in onnx2ncnn at padding op.


wincle:
And implement reflect padding, have tested the code by myself.

nihui:
thanks for your contribution

PR #1119, opened 2019-07-17
Fix protobuf cmake variables

daquexian:
According to cmake doc, FindProtobuf.cmake defines Protobuf_* instead of PROTOBUF_* in cmake >= 3.6

What's more, it's more proper to use Protobuf_INCLUDE_DIRS instead of Protobuf_INCLUDE_DIR according to the doc, the latter one is expected to be set by the user and even doesn't exist in cmake 3.0, so I fixed it too.

Refer: [cmake 3.5 doc](https://cmake.org/cmake/help/v3.5/module/FindProtobuf.html)
[cmake 3.6 doc](https://cmake.org/cmake/help/v3.6/module/FindProtobuf.html)
[cmake 3.0 doc](https://cmake.org/cmake/help/v3.0/module/FindProtobuf.html)


nihui:
The current cmake script is compatible with both old and the latest cmake.
If you encounter build issue, try upgrading cmake.

daquexian:
@nihui how about the INCLUDE_DIRS vs INCLUDE_DIR problem?

PR #1122, opened 2019-07-18
fix #1053

### LINKED ISSUE #1053: feature map size is  subtracted by 2 after Deconvolution layer in ncnn. How to avoid this?
When I convert pytorch model to ncnn, the featrue map was subtracted by 2. That is because the deconvolution layer of pytorch have the parameter _output_padding_ to maintain feature maps' original size. Dose ncnn have  any way to keep the original size when do the deconvolution?

wincle:
Actually it fix my problem, hope the code is right.
I think apply output_padding is just to adjust the out_w and out_h.
If I'm wrong , pls point it out.
Ps: And fix instance norm in ncnnoptimize.

### DIFF
--- a/src/layer/arm/deconvolution_arm.cpp
+++ b/src/layer/arm/deconvolution_arm.cpp
@@ -132,11 +132,11 @@ int Deconvolution_arm::forward(const Mat& bottom_blob, Mat& top_blob, const Opti
     int h = bottom_blob.h;
     size_t elemsize = bottom_blob.elemsize;
 
-    int outw = (w - 1) * stride + kernel_size;
-    int outh = (h - 1) * stride + kernel_size;
+    int outw = (w - 1) * stride + kernel_size + output_pad_w;
+    int outh = (h - 1) * stride + kernel_size + output_pad_h;
 
     Mat top_blob_bordered;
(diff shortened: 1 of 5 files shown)


nihui:
output_padding : zero-padding added to one side of the output

I think the padded value will be the bias value if you only adjust outw outh

wincle:
> output_padding : zero-padding added to one side of the output
> 
> I think the padded value will be the bias value if you only adjust outw outh

Thx ! Checked the code again and you are right~
Any hint about how to implement it in an elegant way?

wincle:
But I think may be added the bias is not unacceptable.

> The padding argument effectively adds dilation * (kernel_size - 1) - padding amount of zero padding to both sizes of the input. This is set so that when a Conv1d and a ConvTranspose1d are initialized with same parameters, they are inverses of each other in regard to the input and output shapes. However, when stride > 1, Conv1d maps multiple input shapes to the same output shape. output_padding is provided to resolve this ambiguity by effectively increasing the calculated output shape on one side. Note that output_padding is only used to find output shape, but does not actually add zero-padding to output.
Quote from https://pytorch.org/docs/stable/nn.html#torch.nn.ConvTranspose2d.
Does the value added to the side really affect?

Issue #1093, opened 2019-07-10
onnx2cnn doesn't export MemoryData properly


arnaudbrejeon:
When exporting MemoryData, the logic doesn't take into account when M.dims_size() is zero and there actually is some data. The consequence is that reading the binary file will be offsetted.

PR #1187, opened 2019-08-13
Fix for onnx2ncnn Div op

Howave:
update for issue #1093


nihui:

PR #1188, opened 2019-08-13
Fix for MemoryData params in ncnn optimize tool

Howave:

PR #1194, opened 2019-08-17
Write necessary data for memorydata in ncnnoptimize.

xunings:
See https://github.com/Tencent/ncnn/pull/1188


nihui:

PR #1195, opened 2019-08-18
Implemented hard swish layer

Howave:
not tested yet


nihui:

PR #1211, opened 2019-08-25
Update ncnn2table.cpp

MambaWong:
add swapRB flag


nihui:

PR #1233, opened 2019-09-03
fix ConvolutionDepthWise int8_requantize

yx9527:
Fix a bug in ConvolutionDepthWise::forward when use_int8_requantize.


nihui:
https://github.com/Tencent/ncnn/blob/c4bebc63714662aa2d7446be1728bb0feca07361/src/layer/convolutiondepthwise.cpp#L407
https://github.com/Tencent/ncnn/blob/c4bebc63714662aa2d7446be1728bb0feca07361/src/layer/convolutiondepthwise.cpp#L470

and this ?

nihui:
@yx9527  请一起修改了吧

nihui:
thanks for your contribution

PR #1262, opened 2019-09-17
onnx2ncnn: add axes support to slice and fix bug 1261

arnaudbrejeon:
Add axes support to slice and fix bug 1261


nihui:
I added the starts, ends, axes attributes to crop
https://github.com/Tencent/ncnn/commit/a6c60068e69c1458f1ad3c707a74239e2ccbdeff

PR #1282, opened 2019-09-30
Fixed broken link

volvet:
Fixed broken link of operation-param-weight-table


nihui:
md suffix is intentionally omitted for compatibility of the wiki webpage
see
https://github.com/Tencent/ncnn/wiki/param-and-model-file-structure#layer-param

PR #1299, opened 2019-10-11
add split op support for onnx2ncnn

bindog:
Some of our models use the [split](https://github.com/onnx/onnx/blob/master/docs/Operators.md#Split) op in onnx, but the onnx2ncnn do not support it, so I implement it according to the related code in caffe2ncnn
https://github.com/Tencent/ncnn/blob/25979c1bdb73d58d0a463abb7742a46c44e30f47/tools/caffe/caffe2ncnn.cpp#L1574-L1601
I test it on two of our models, and it works fine. But I'm not sure if it cover all the edge cases, please review it @nihui


nihui:

PR #1302, opened 2019-10-14
fix opencv compatibility issue with major verion > 2 for the quantisation tool

ShuangLiu1992:
fix opencv compatibility issue with major verion > 2 for the quantisation tool


nihui:
thanks for your contribution！

PR #1307, opened 2019-10-16
Add assembly int8 gemm

tpoisonooo:
1. use assembly arm64 symmetric int8 GEMM to implement conv1x1 and im2col convolution. please check [chgemm](https://github.com/tpoisonooo/chgemm);
2. split `convolution_arm.cpp` into  3 parts;
3. test with part of squeezenet, compile with `-O3` flag.

test conv1x1, loop 1000 times:

| old version | this version | speedup |
| :---: | :---: | :---: |
|1.631s | 1.190s | 27%|

use im2col convolution to run all layers, repeat hundreds of times:

| old version | this version | speedup |
| :---: | :---: | :---: |
|7.548s | 6.814s | 9.7%|

4. test full squeezenet1.1, repeat 21 times, single thread, no requant shall be allowed.

>tpoisonooo@RK3399:~/ncnn/build/examples$ time ./squeezenet ~/Desktop/227.jpg
find_blob_index_by_name data failed
639 = 0.026630
837 = 0.026277
836 = 0.020731
real	0m1.801s
user	0m1.747s
sys	0m0.050s

| old version average timecost | this version | speedup |
| :---: | :---: |  :---: |
| 81ms| 73ms | 9.8% |

5. since iOS cross compile environment is really hard for me,  iOS arm64 still use old code, let it go~ 

Appendix:  part_squeezenet.param
>tpoisonooo@RK3399:~/ncnn/build/tools/caffe$ cat part_squeezenet.param
7767517
8 9
Convolution      conv1            1 1 data conv1 0=64 1=3 2=1 3=2 4=0 5=1 6=1728 8=2
ReLU             relu_conv1       1 1 conv1 conv1_relu_conv1
Pooling          pool1            1 1 conv1_relu_conv1 pool1 0=0 1=3 2=2 3=0 4=0
Convolution      fire2/squeeze1x1 1 1 pool1 fire2/squeeze1x1 0=16 1=1 2=1 3=1 4=0 5=1 6=1024 8=2
ReLU             fire2/relu_squeeze1x1 1 1 fire2/squeeze1x1 fire2/squeeze1x1_fire2/relu_squeeze1x1
Convolution      fire2/expand1x1  1 1 fire2/squeeze1x1_fire2/relu_squeeze1x1 fire2/expand1x1 0=64 1=1 2=1 3=1 4=0 5=1 6=1024 8=2
ReLU             fire2/relu_expand1x1 1 1 fire2/expand1x1 fire2/expand1x1_fire2/relu_expand1x1
Softmax          prob             1 1 fire2/expand1x1_fire2/relu_expand1x1 prob 0=0 1=1


src/layer/arm/gemm/int8kernel_m1.S
```diff
@@ -0,0 +1,515 @@
...
+    stp d10, d11, [sp, #(1 * 16)]
+    stp d12, d13, [sp, #(2 * 16)]
+    stp d14, d15, [sp, #(3 * 16)]
+    stp d16, d17, [sp, #(4 * 16)]
+    stp x18, x19, [sp, #(5 * 16)]
+    stp x20, x21, [sp, #(6 * 16)]
+    stp x22, x23, [sp, #(7 * 16)]
+    stp x24, x25, [sp, #(8 * 16)]
+    stp x26, x27, [sp, #(9 * 16)]
+    str x28, [sp, #(10 * 16)]
```
nihui: The ARM 64-bit architecture also has a further thirty-two registers, v0-v31, which can be used by SIMD and Floating-Point operations. The precise name of the register will change indicating the size of the access.

Note: Unlike in AArch32, in AArch64 the 128-bit and 64-bit views of a SIMD and Floating-Point register do not overlap multiple registers in a narrower view, so q1, d1 and s1 all refer to the same entry in the register bank.

The first eight registers, v0-v7, are used to pass argument values into a subroutine and to return result values from a function. They may also be used to hold intermediate values within a routine (but, in general, only between subroutine calls).

Registers v8-v15 must be preserved by a callee across subroutine calls; the remaining registers (v0-v7, v16-v31) do not need to be preserved (or should be preserved by the caller). Additionally, only the bottom 64-bits of each value stored in v8-v15 need to be preserved; it is the responsibility of the caller to preserve larger values.

reference
https://stackoverflow.com/questions/261419/what-registers-to-save-in-the-arm-c-calling-convention/29603040#29603040

tpoisonooo:
Fixed.

tpoisonooo:
int8 GEMM would rewrite with inline assembly later, do not merge.
Unit test is also needed.

tpoisonooo:
All GEMM test case has passed, pls review.

src/layer/arm/gemm_symm_int8.h
```diff
@@ -0,0 +1,2636 @@
+#if __aarch64__
```
nihui: copyright header is missing

PR #1308, opened 2019-10-16
[WIP] add reduce op support for onnx

bindog:
There are a lot of reduce op in onnx(https://github.com/onnx/onnx/blob/master/docs/Operators.md)
I implement and extend it based on `src/layer/reduction.cpp`, but I'm not familiar with some of the functions in ncnn, so the usage of function or code style may not consistent with this repo...

So please help review and test this PR, and give some advices~ @nihui

### DIFF
--- a/src/layer/reduction.cpp
+++ b/src/layer/reduction.cpp
@@ -782,7 +782,7 @@ int Reduction::forward(const Mat& bottom_blob, Mat& top_blob, const Option& opt)
     */
     
     int dims = bottom_blob.dims;
-    int axes_flag[dims] = {0};
+    int axes_flag[3] = {0};
     int reduction_type = -233;
 
     if (type)
--- a/src/layer/reduction.cpp
+++ b/src/layer/reduction.cpp
@@ -34,12 +34,13 @@ int Reduction::load_param(const ParamDict& pd)
     operation = pd.get(0, 0);
     dim = pd.get(1, 0);
     coeff = pd.get(2, 1.f);
+    keepdims = pd.get(3, 0);
 
     return 0;
 }
 
 template<typename Op, typename Op2>
-static int reduction_op(const Mat& a, Mat& b, float v0, int dim, float coeff, const Option& opt)
+static int reduction_op(const Mat& a, Mat& b, float v0, int dim, float coeff, int keepdims, const Option& opt)
 {
     Op op;
     Op2 op2;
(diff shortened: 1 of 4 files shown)


nihui:
非常感谢w

currently, ncnn Reduction can only handle 3-dim blob, and never keepdims
the dim attribute of ncnn Reduction is actually a reduction type, not the axes of onnx Reduce

1 rename the attribute name "dim" to "type"
2 add the axes attribute to ncnn Reduction, default to empty for backward compability
3 add the keepdims attribute to ncnn Reduction, default to 0 for backward compability
4 implement 2-dim and 1-dim blob reduction
5 implement keepdims version of all reduction types
6 select the proper reduction type according to the ("axes" or "type") attribute AND the bottom_blob dims rank
7 onnx opeartor conversion

do not mix different reduction version into one, for example keepdims and non-keepdims, or 3-dim and 2-dim reduction
don't be scare of lots of "if" conditions and branches
you may refer how softmax implements many variants for different blob dims and axis
keep it simple and stupid for naive implementation, clever code goes to x86/arm/...

bindog:
I implement most of the requirements above, but there still remain some issues need further discussion

1. the design of `type` and `reduction_type`, and here is the details in comments
https://github.com/bindog/ncnn/blob/308dbca9a62cec88b7a0b2354bd604084ca7570c/src/layer/reduction.cpp#L771-L782
2. the `ReduceL2` `ReduceLogSum` and `ReduceLogSumExp` need post process after reduce operation(`sqrt` and `log`), so I add a `post_process` template for them 
3. currently, the `coeff` parameter is not used by onnx, I use it in `ReduceMean` to avoid another post process calculation, but I'm afraid it will cause conflict with the usage with caffe or some precision loss maybe... please help confirm it

there may be some other issue which not consistent with the style of ncnn(both in design and code style), please help check and review this, thanks~ @nihui

nihui:
1. rename type to reduce_all, with more obvious attribute meaning

2. ah, it seems the all-in-one reduction_type is quite messy, it's my fault
prefer splitting into three bool reduce_w, reduce_h and reduce_c

```
if (dims == 3)
{
    if (reduce_w && reduce_h && !reduce_c)
    {
        // ...
    }

    if (reduce_w && reduce_h && reduce_c)
    {
        // ...
    }

    // ... other reduce types
}
```

3. the reduction_op template function is too long now
move the keepdims block into another template function, maybe reduction_op_keepdims

4. rename post_process to reduction_post_process
it is an element-wise unary function
actually, the coeff multiplication is a post process too, and coeff is usually 1.f
move the coeff multiplication into reduction_post_process, so we don't pay for this multiplication during reduction if it equals to 1

bindog:
changes updated~

nihui:
非常感谢w

PR #1326, opened 2019-10-25
fix Crop bugs

hegc:
bug code : if start=0 and end =1, 
                       _woffset will be equal to value of w
                       _outw will be equal to 1 - w
if value of _outw is less than 0, code will be dump!

right code: if start =0, _woffset will be 0!

Please fix this bug.

### DIFF
--- a/src/layer/arm/crop_arm.cpp
+++ b/src/layer/arm/crop_arm.cpp
@@ -120,19 +120,19 @@ int Crop_arm::forward(const Mat& bottom_blob, Mat& top_blob, const Option& opt) 
 
             if (dims == 1) // axis == 1
             {
-                _woffset = start > 0 ? start : w * elempack + start;
+                _woffset = start >= 0 ? start : w * elempack + start;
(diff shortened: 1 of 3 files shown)


tencent-adm:
[CLA assistant check](https://cla.opensource.tencent.com/Tencent/ncnn?pullRequest=1326) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.opensource.tencent.com/Tencent/ncnn?pullRequest=1326) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.opensource.tencent.com/check/Tencent/ncnn?pullRequest=1326) it.</sub>

nihui:
hi，thanks

and please fix for https://github.com/Tencent/ncnn/blob/master/src/layer/arm/crop_arm.cpp and https://github.com/Tencent/ncnn/blob/master/src/layer/vulkan/crop_vulkan.cpp

nihui:
thanks for your contribution！

PR #1328, opened 2019-10-25
[WIP] add all kinds of broadcasting type for binaryop

bindog:
This is an alpha version of binaryop which support all kinds of broadcasting type, I put some of the details in docs.  Please review the design and code style of this PR first,  and I will run more test cases this weekend to make sure it will cover all edge cases... @nihui 

If you think the design is OK, then I will try to implement the ARM version

### DIFF
--- a/docs/developer-guide/binaryop-broadcasting.md
+++ b/docs/developer-guide/binaryop-broadcasting.md
@@ -6,30 +6,43 @@ C = BinaryOp(A, B)
 
 shape notation convention is [w], [w,h], [w,h,c]
 
+Two tensors are “broadcastable” if the following rules hold:
+
+- Each tensor has at least one dimension.
+- When iterating over the dimension sizes, starting at the last dimension, the dimension sizes must either be equal, one of them is 1, or one of them does not exist.
+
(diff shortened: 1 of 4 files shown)


nihui:
The AND operator inside loop is a bit ugly and maybe slower if they are at the critical inner-est loop
But I haven't figure out a style for both straight-forward and less code ...  :|
Anyway, that's a naive implementation

```
outptr[j] = op(ptr[j & wf], ptr1[j & wf1]);
```

ow -> outw, oh -> outh, oc -> outc
attribute name layout is ambiguous as ncnn Mat memory layout is always C-H-W, rename to numpy_style
function name can be binary_op_numpy for indicating the numpy broadcasting style

You don't need to adopt the same rules to the existing ncnn style broadcasting types, the old code block can be leaved unchanged

daichuanliang:
This change  only supports x86, it doesn't support arm.

nihui:
implemented in https://github.com/Tencent/ncnn/commit/ab4cfbf5b05550f38f6aa562c1e583fa92d73da3

Issue #1332, opened 2019-10-27
POSITION_INDEPENDENT_CODE


ShuangLiu1992:
Can we please not have POSITION_INDEPENDENT_CODE turned on and consequently set INTERFACE_POSITION_INDEPENDENT_CODE to be true by default for non MSVC compilers? some compiler and setups might not want to have their compiler flags poisoned with PIC when linking ncnn.

```
if(CMAKE_CXX_COMPILER_ID MATCHES "MSVC"
    OR (CMAKE_CXX_COMPILER_ID MATCHES "Clang" AND CMAKE_CXX_SIMULATE_ID MATCHES "MSVC"))
    target_compile_definitions(ncnn
        PRIVATE _SCL_SECURE_NO_WARNINGS _CRT_SECURE_NO_DEPRECATE
        PUBLIC NOMINMAX)
else()
    target_compile_options(ncnn
        PRIVATE -Wall -Wextra -Wno-unused-function)
    set_target_properties(ncnn
        PROPERTIES
            POSITION_INDEPENDENT_CODE ON
            INTERFACE_POSITION_INDEPENDENT_CODE ON)
```

ShuangLiu1992:
if a ncnn user wants to enable POSITION_INDEPENDENT_CODE I'm pretty sure they can add -DCMAKE_POSITION_INDEPENDENT_CODE=ON at config time themselves...

Or, since delivering ncnn release binaries to build Android shared library requires PIC must to be on, could we please have an option to turn it off if for some specific uses where people don't want PIC?

nihui:
merged https://github.com/Tencent/ncnn/pull/1333

PR #1333, opened 2019-10-27
position independent code

ShuangLiu1992:
https://github.com/Tencent/ncnn/issues/1332


nihui:
Thanks for your contribution!

Issue #1339, opened 2019-10-28
NetOptimize::fuse_convolutiondepthwise_batchnorm


ShuangLiu1992:
NetOptimize::fuse_convolutiondepthwise_batchnorm seems to produce incorrect results (regardless of whether group number is the same as input output channels), NetOptimize::fuse_convolution_batchnorm works as expected.

ShuangLiu1992:
more specifically, it is depthwise convolution with bias terms followed by batch norm. I'm retraining a model to see if dep con without bias produce the same result. but looking at ncnnoptimize it shouldn't be.

ShuangLiu1992:
dep convolution without bias seems to not suffer from this, which is very weird... not sure if it's my training setup or something going wrong during conversion. I will retrain to see if I can reproduce this issue and share the model if I have time to retrain a couple of times....

nihui:
change
```
                bias[i] += a[i];
```
to
```
                bias[i] = bias[i] * b[i] + a[i];
```

ShuangLiu1992:
hmmm, should I do that for regular conv as well?

nihui:
apply to all conv/deconv/innerproduct - batchnorm family
Because in most cases, we usually disable the convolution bias when it is followed by batchnorm, so it may not be such serious, but it is definitely a bug ...

nihui:
merged https://github.com/Tencent/ncnn/pull/1340

PR #1340, opened 2019-10-28
fix ncnnoptimize bias term bug #1339

ShuangLiu1992:
fix ncnnoptimize bias term bug #1339
https://github.com/Tencent/ncnn/issues/1339


ShuangLiu1992:
fix ncnnoptimize bias term bug https://github.com/Tencent/ncnn/issues/1339

PR #1344, opened 2019-10-29
fix errors: spp::copy_make_border and std:min/max

CosmosHua:
1.fix error in spp::copy_make_border  2.fix errors of std:min/max reported on VS2015

### DIFF
--- a/src/allocator.cpp
+++ b/src/allocator.cpp
@@ -430,7 +430,7 @@ VkBufferMemory* VkBlobBufferAllocator::fastMalloc(size_t size)
         }
     }
 
-    size_t new_block_size = std::max(block_size, aligned_size);
+    size_t new_block_size = (std::max)(block_size, aligned_size);
 
     // create new block
     VkBufferMemory* block = new VkBufferMemory;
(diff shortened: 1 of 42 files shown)


tencent-adm:
[CLA assistant check](https://cla.opensource.tencent.com/Tencent/ncnn?pullRequest=1344) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.opensource.tencent.com/Tencent/ncnn?pullRequest=1344) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.opensource.tencent.com/check/Tencent/ncnn?pullRequest=1344) it.</sub>

nihui:
ncnn defines NOMINMAX macro for msvc in CMakeLists.txt, so if you use cmake to generate the vs project, I don't think std::min/max error will happen.
please provide more info about your building steps, thanks !

https://github.com/Tencent/ncnn/blob/2f26697aea3de6965879e57b897ee1f18d1e3826/src/CMakeLists.txt#L307

CosmosHua:
Because I omit to compile protobuf and compile ncnn directly using cmake, then copy the generated *.h to src/. Then I create a new project in VS2015 and compile src/. I have tried to add #define NOMINMAX before any #include statement in my cpp, but useless. So I bracket std::max/min, then it works.
As you said, it raise errors because I miss the sucess compilation of ncnn using cmake.

nihui:
You can also try cmake-gui to generate a proper vs2015 project
I closed PR as the issue is due to build steps.

PR #1345, opened 2019-10-29
fix spp::copy_make_border

CosmosHua:
fix error in spp calling copy_make_border.


nihui:

PR #1346, opened 2019-10-29
Fix min/max macro issue

akemimadoka:
This patch makes headers of NCNN can be used without bothering by min/max macro issue even not introduced by CMake.


nihui:
define NOMINMAX for benchncnn.cpp ?

PR #1347, opened 2019-10-30
fix negative axis in concat

bindog:
When I debug a model for my friend, I found negative axis in his concat op...
https://github.com/onnx/onnx/blob/master/docs/Operators.md#concat

### DIFF
--- a/src/layer/concat.h
+++ b/src/layer/concat.h
@@ -29,7 +29,7 @@ public:
     virtual int forward(const std::vector<Mat>& bottom_blobs, std::vector<Mat>& top_blobs, const Option& opt) const;
 
 public:
-    int _axis;
+    int axis;
 };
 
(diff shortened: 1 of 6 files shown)


src/layer/concat.h
```diff
@@ -29,7 +29,7 @@ class Concat : public Layer
     virtual int forward(const std::vector<Mat>& bottom_blobs, std::vector<Mat>& top_blobs, const Option& opt) const;
 
 public:
-    int axis;
+    int _axis;
```
nihui: please do not prefix with underscore for attribute identifier

bindog:
Sorry, I will learn some vulkan basic before I can fix this

nihui:
move to https://github.com/Tencent/ncnn/pull/2365

PR #1357, opened 2019-11-05
fix bug that MemoryData as multi operations input

GuoxiaWang:
When I convert tensorflow pb to onnx format model firstly, I meet  MemoryData as multi operations input problem when I using `onnx2ncnn` tool convert onnx model to ncnn model. I double check there are miss a `Split` operation for MemoryData type if it as multi operations input.

### DIFF
--- a/tools/quantize/ncnn2table.cpp
+++ b/tools/quantize/ncnn2table.cpp
@@ -77,12 +77,31 @@ public:
     int get_conv_names();
     int get_conv_bottom_blob_names();
     int get_conv_weight_blob_scales();
+    int get_input_names();
 
 public:
     std::vector<std::string> conv_names;
     std::map<std::string,std::string> conv_bottom_blob_names;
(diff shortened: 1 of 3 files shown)


tools/quantize/ncnn2table.cpp
```diff
@@ -77,13 +77,32 @@ class QuantNet : public ncnn::Net
     int get_conv_names();
     int get_conv_bottom_blob_names();
     int get_conv_weight_blob_scales();
+    int get_input_names();
```
nihui: get_input_names() defined but never called ?

GuoxiaWang: Sorry, It need be called before `ex.input`. In my other local repository, I add before 
```
    net.get_conv_names();
    net.get_conv_bottom_blob_names();
    net.get_conv_weight_blob_scales();
```

I fix it immediately.

PR #1363, opened 2019-11-08
fix Reduction Op attr bug

GuoxiaWang:


tools/ncnnoptimize.cpp
```diff
@@ -2230,6 +2230,8 @@ int NetOptimize::save(const char* parampath, const char* binpath)
             fprintf_param_value(" 0=%d", operation)
             fprintf_param_value(" 1=%d", reduce_all)
             fprintf_param_value(" 2=%e", coeff)
+            { if (!op->axes.empty()) fprintf_param_int_array(3, op->axes, pp); }
+            fprintf_param_value(" 4=%e", keepdims)
```
nihui: %d

tools/quantize/ncnn2int8.cpp
```diff
@@ -993,6 +993,8 @@ int NetQuantize::save(const char* parampath, const char* binpath)
             fprintf_param_value(" 0=%d", operation)
             fprintf_param_value(" 1=%d", reduce_all)
             fprintf_param_value(" 2=%f", coeff)
+            { if (!op->axes.empty()) fprintf_param_int_array(3, op->axes, pp); }
+            fprintf_param_value(" 4=%e", keepdims)
```
nihui: %d

nihui:
Thanks for your contribution  ！

PR #1410, opened 2019-11-29
Fix ncnn2table coredump

Mycaster:
I download a mobilenet_v2 model from https://github.com/shicai/MobileNet-Caffe , and try to quantize it.  but [Input] layer is not defined in mobilenet_deploy.prototxt. 

when  [Input] layer not found in ncnn.praram, ncnn2table will throw Segment Fault exception, which caused by the following code: 

```
+664         ncnn::Extractor ex = net.create_extractor();
+665         ex.input(net.input_names[0].c_str(), in);
```
there will be better to check the input_names.size()  before quantize


nihui:

PR #1417, opened 2019-12-08
Kill tons of warnings.

ufoym:


nihui:
please fix the code instead of ignoring via compilation flags
you can refer how https://github.com/Tencent/ncnn/pull/1286 does

PR #1428, opened 2019-12-16
Fix warnings on Visual Studio

chosungmann:


src/layer/padding.cpp
```diff
@@ -298,7 +298,7 @@ int Padding::forward(const Mat& bottom_blob, Mat& top_blob, const Option& opt) c
             return -100;
 
         if (elemsize == 1)
-            copy_make_border_image<signed char>(bottom_blob, top_blob, 0, left, type, value);
+            copy_make_border_image(bottom_blob, top_blob, 0, left, type, static_cast<signed char>(value));
```
nihui: keep the explicit function template arg type here, and the following changes too
```
copy_make_border_image<signed char>(bottom_blob, top_blob, 0, left, type, static_cast<signed char>(value));
```

chosungmann: done

nihui:

PR #1430, opened 2019-12-17
Add mips architecture optimization support

nullptr-leo:
When architecture set to mips, try to check and compile layers in `src/layer/mips` directory.


pineking:
there are no files in `src/layer/mips`,  do you forget to add them?

nullptr-leo:
> there are no files in `src/layer/mips`, do you forget to add them?

This commit only contains changes for compiling support.

nihui:
It seems to be useless as there is no mips folder actually though, but it is not too bad

PR #1439, opened 2019-12-18
Replace all occurrences of std::{unary|binary}_function<> with std::function<>

chosungmann:
std::{unary|binary}_function<> are deprecated in C++11 and removed in C++17.

Actually, these were unnecessary in general. In the C++98/03 era, many user-
defined function object classes derived from these base classes in an attempt
to imitate STL conventions. However, STL containers and algorithms have never
required such inheritance (or the typedefs that they provide). Only the
function object adaptors (like bind1st()) needed such typedefs. Currently,
there are no uses of function object adapters in our codebase, so we can
eliminate the inheritance completely.

### DIFF
--- a/src/layer/arm/binaryop_arm.cpp
+++ b/src/layer/arm/binaryop_arm.cpp
@@ -519,22 +519,22 @@ static int binary_op_scalar_inplace(Mat& a, float b, const Option& opt)
 }
 
 template<typename T>
-struct binary_op_add : std::binary_function<T,T,T> {
+struct binary_op_add {
     T operator() (const T& x, const T& y) const { return vaddq_f32(x, y); }
 };
 
 template<typename T>
-struct binary_op_sub : std::binary_function<T,T,T> {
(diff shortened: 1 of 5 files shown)


nihui:
Hi, we don't plan to enable c++11 features for the ncnn library codebase for maximum portablibity.
As far as I know, quite many developers port ncnn library to thier embedded platforms with VERY OLD toolchains.

If you think this change is critical or may bring runtime speedup, please add a compile-time cmake option for all c++11 features, and guard modern code with macros conditions.

chosungmann:
Then why don't we eliminate the inheritances of `std::{unary|binary}_function<>`? If do so, we can support MODERN toolchains as well as VERY OLD toolchains.

nihui:
> Then why don't we eliminate the inheritances of `std::{unary|binary}_function<>`? If do so, we can support MODERN toolchains as well as VERY OLD toolchains.

Ah, that is perfect !

PR #1505, opened 2020-01-19
fix quantize tools cross platform issues

kalcohol:
1. for cross platform, use cv::CommandLineParser instead of getopt.h
2. use cv::glob to search folder instead of dirent.h
3. fix some other warnings of function from cstdlib
4. add some const and some auto

now the tools is usable.

### DIFF
--- a/tools/quantize/ncnn2int8.cpp
+++ b/tools/quantize/ncnn2int8.cpp
@@ -688,9 +688,7 @@ int NetQuantize::save(const char* parampath, const char* binpath)
             fprintf_param_value(" 6=%d", woffset2)
             fprintf_param_value(" 7=%d", hoffset2)
             fprintf_param_value(" 8=%d", coffset2)
-            {
-                if (!op->starts.empty()) fprintf_param_int_array(9, op->starts, pp);
-            }
+            { if (!op->starts.empty()) fprintf_param_int_array(9, op->starts, pp); }
             { if (!op->ends.empty()) fprintf_param_int_array(10, op->ends, pp); }
             { if (!op->axes.empty()) fprintf_param_int_array(11, op->axes, pp); }
         }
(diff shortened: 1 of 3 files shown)


tencent-adm:
[CLA assistant check](https://cla.opensource.tencent.com/Tencent/ncnn?pullRequest=1505) <br/>All committers have signed the CLA.

kalcohol:
oh no, all CI do not has opencv......

tools/quantize/ncnn2int8.cpp
```diff
@@ -505,16 +541,16 @@ int NetQuantize::save(const char* parampath, const char* binpath)
         ncnn::ParamDict pd;
         layer_default->load_param(pd);
 
-#define fprintf_param_value(format, phase) \
+#define FPRINTF_PARAM_VALUE(format, phase) \
```
nihui: keep the lowercase style macro

kalcohol: solved.

kalcohol:
why still show 1 change requested...

nihui:
keep the one line style block

```
            { if (!op->starts.empty()) fprintf_param_int_array(9, op->starts, pp); }
```
```
            {
                if (!op->starts.empty()) fprintf_param_int_array(9, op->starts, pp);
            }
```

pineking:
we use quantize tools without opencv, that's not a good idea for using these opencv dependencies.

kalcohol:
> we use quantize tools without opencv, that's not a good idea for using these opencv dependencies.

um, which image decoder did u chosen?

pineking:
> > we use quantize tools without opencv, that's not a good idea for using these opencv dependencies.
> 
> um, which image decoder did u chosen?

ignore it, that's my mistake, for offline tools, such as quantize, benchmark. actually, we also use opencv to read image files.

PR #1563, opened 2020-02-28
Add badges for readme

kalcohol:
close it if u do not like it


nihui:
It is TOO messy and complicated

PR #1566, opened 2020-02-29
add skip cooling down option

kalcohol:
add option for cooling down
add the option usage
add benchmark of RTX2080


nihui:

PR #1578, opened 2020-03-04
test_tanh.cpp: Add unittest of tanh

songqun:
直接仿照test_sigmoid.cpp了


nihui:

PR #1590, opened 2020-03-06
fix requant relu6 bug

xieydd:
As the title, just change the backend x86.

### DIFF
--- a/src/layer/clip.cpp
+++ b/src/layer/clip.cpp
@@ -16,6 +16,14 @@
 
 #include <float.h>
 
+static inline signed char float2int8(float v)
+{
+    int int32 = static_cast<int>(round(v));
+    if (int32 > 127) return 127;
+    if (int32 < -127) retuurn -127;
+    return (signed char)int32;
+}
+
 namespace ncnn {
 
 DEFINE_LAYER_CREATOR(Clip)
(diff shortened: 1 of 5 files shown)


nihui:
```
ncnn::ParamDict pd;
pd.set(0, activation_params[0] * top_blob_int8_scale);// min
pd.set(1, activation_params[1] * top_blob_int8_scale);// max
activation->load_param(pd);
```

Move these loading stuff of activation to create_pipeline()

All pipelines should never be changed inside forward/forward_inplace, that would cause data race in multithreading application.

xieydd:
> ```
> ncnn::ParamDict pd;
> pd.set(0, activation_params[0] * top_blob_int8_scale);// min
> pd.set(1, activation_params[1] * top_blob_int8_scale);// max
> activation->load_param(pd);
> ```
> 
> Move these loading stuff of activation to create_pipeline()
> 
> All pipelines should never be changed inside forward/forward_inplace, otherwise it would cause data race in multithreading application.

`use_intt8_requant ` is set in confuse_networl after every layer create_pipeline? i am a little confuse it.about

nihui:
> 
> 
> > ```
> > ncnn::ParamDict pd;
> > pd.set(0, activation_params[0] * top_blob_int8_scale);// min
> > pd.set(1, activation_params[1] * top_blob_int8_scale);// max
> > activation->load_param(pd);
> > ```
> > 
> > 
> > Move these loading stuff of activation to create_pipeline()
> > All pipelines should never be changed inside forward/forward_inplace, otherwise it would cause data race in multithreading application.
> 
> `use_intt8_requant ` is set in `confuse_network` after every layer create_pipeline? i am a little confuse about it.
> 
> Can i change the order of `load_model create_pipeline fuse_nettwork ` to `load_model fuse_nettwork create_pipeline`?

The confuse_network function is actually a legacy issue, and may probably removed in the future.
But for now, I think it is fine to adjust this order or make changes to it, as long as it could work.

xieydd:
@nihui Temporary solution.

nihui:
OK, but that's it

PR #1593, opened 2020-03-07
test_clip.cpp: Add unittest of clip

xieydd:
As the title.


nihui:
RandomMat fills the mat with values ranging -2 ~ 2, use 6 as the clip max may not proper here

xieydd:
@nihui , i see, change 6 to 2 is ok? or change the RandomMat is better?

nihui:
Mat a = RandomMat(100, 100, 16)
Randomize(a, -10.f, 10.f);

nihui:

PR #1597, opened 2020-03-07
test_hardsigmoid.cpp: Add unittest of HardSigmoid

xieydd:
Same as the title.


nihui:

PR #1598, opened 2020-03-08
test_hardswish.cpp: Add unittest of HardSwish

xieydd:
Same as the title.


nihui:

PR #1622, opened 2020-03-17
add build script for building android lib on windows via ndk

kalcohol:
as the title


nihui:
rename build.cmd to build-android.cmd

PR #1624, opened 2020-03-18
fix layer innerproduct when build with requant option on

yx9527:


nihui:

PR #1633, opened 2020-03-19
fix aarch64 build

JackPumpkins:
Use the Special Format String %= as defined [here.](https://gcc.gnu.org/onlinedocs/gcc/Extended-Asm.html#AssemblerTemplatel) to avoid duplicate definition error during the compilation


nihui:
use number as label

JackPumpkins:
replace label by number

nihui:

PR #1691, opened 2020-04-13
Tiny fix mxnet2ncnn tool

mengfu188:
Hi nihui.
Mxnet slice step parameter is option, if not given in model code, it will ignore in mxnet-symbol.json. So i add check.


nihui:

PR #1730, opened 2020-04-28
Bugfix in crop.cpp

ddddwee1:
1. Mistakenly used _w_offset for _outh
2. When converting MXNet slice layer to NCNN crop layer, the dimension with "None" value in MXNet will convert to "-233" but it was not handled by NCNN crop layer. It will cause error.


nihui:

PR #1733, opened 2020-04-30
Add mish layer

zhiliu6:


nihui:
what about adding a test for this new operator  :P

nihui:

PR #1740, opened 2020-05-06
Add darknet to ncnn converter with YOLOv4 support.

zhiliu6:


nihui:
please fix the ci error

nihui:
add your copyright info in darknet2ncnn.cpp

nihui:
NICE!

nihui:
GREAT JOB!

AlexeyAB:
So does NCNN now support YOLOv4? It isn't specified in the Readme.

PR #1741, opened 2020-05-07
mxnet2ncnn support list args in attrs

papercatnku:
Either tuple or list could be valid argumenst in mxnet.gluon.nn.Conv* classes. 
current mxnet2ncnn tool failed to work when set kernel/pad/stride/dialate with list instead of tuple.


nihui:

PR #1743, opened 2020-05-08
darknet2ncnn: Add merge output option.

zhiliu6:


nihui:

PR #1752, opened 2020-05-12
Optimize Mish vulkan performance

zhiliu6:


nihui:
Thanks for your contribution

PR #1761, opened 2020-05-13
Add fused convolution and mish layer support.

zhiliu6:
with vulkan, x86 and arm support


zhiliu6:
> 
> 
> NCNN_vulkan_fuse_conv_mish macro is redundant

removed

nihui:
please fix the unittest error when arm packing and bf16s enabled

nihui:
YEAH !

PR #1766, opened 2020-05-15
fix ncnn2mem.cpp param bin save path

GuoxiaWang:
fix ncnn2mem.cpp param bin save path


nihui:

PR #1768, opened 2020-05-16
add statisticspooling layer

Wang-Charles:
I'm trying to deploy my **kaldi** model to  mobile platform and almost successful.  In the process, Statistics Pooling layer is added.

Statistics Pooling layer is a layer in **TDNN**, it is used to reduction features from frame level to segment level (_Deep Neural Network Embeddings for Text-Independent Speaker Verification, Dan Povey, interspeech 2017_). In **kaldi** when we using **X-vector system**, we offen get a model with this spicial layer. It is just the mean across all frames (if set include_stddev=true, also include the standard deviation). 

I also find the **reduction.cpp** and **reduction.h** in source code, but I can not find the standard deviation implementation.


src/layer/statisticspooling.cpp
```diff
@@ -0,0 +1,80 @@
...
+// under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+// CONDITIONS OF ANY KIND, either express or implied. See the License for the
+// specific language governing permissions and limitations under the License.
+
+#include "statisticspooling.h"
+#include <float.h>
+#include <limits.h>
+#include <math.h>
+#include <algorithm>
+#include <functional>
```
nihui, line 20: It seems these headers are not needed

```
#include <algorithm>
#include <functional>
```

src/layer/statisticspooling.cpp
```diff
@@ -0,0 +1,80 @@
+// Tencent is pleased to support the open source community by making ncnn available.
+//
+// Copyright (C) 2017 THL A29 Limited, a Tencent company. All rights reserved.
```
nihui, line 3: please update your info in license header

src/layer/statisticspooling.h
```diff
@@ -0,0 +1,38 @@
+// Tencent is pleased to support the open source community by making ncnn available.
+//
+// Copyright (C) 2017 THL A29 Limited, a Tencent company. All rights reserved.
```
nihui, line 3: please update your info in license header

nihui:
If you don't intend to exclude this new layer from ncnn library, you need to add ncnn_add_layer entry in src/CMakeLists.txt

src/layer/statisticspooling.cpp
Wang-Charles: Thanks for your suggestions，I delete these headers.

src/CMakeLists.txt
```diff
@@ -221,6 +221,7 @@ ncnn_add_layer(Power)
 ncnn_add_layer(PReLU)
 ncnn_add_layer(Proposal)
 ncnn_add_layer(Reduction)
+ncnn_add_layer(StatisticsPooling)
```
nihui, line 224: new layer should always be appended at the end of this list
Because the layer enum index is generated from this order. Inserting would break binary compatibility.

Wang-Charles: OK, I will modify that.

Wang-Charles: Do I need to recommit a version.

nihui:

PR #1820, opened 2020-06-06
Improve ROIAlign (accelerate ROIAlign, support sampling ratio and aligned ROIAlign)

wkcn:
Hi there,
In this PR, the code of ROIAlign is adapted from [detectron2 ROIAlign](https://github.com/facebookresearch/detectron2/blob/master/detectron2/layers/csrc/ROIAlign/ROIAlign_cpu.cpp).

It supports sampling ratio and aligned ROIAlign.
ROIAlign with `sampling_ratio=0 and aligned=False` is consistent with the eldder version.

Besides, it pre-calculates the sampling weight to accelerate this operator.

Benchmark on unittest:
before|after
---|---
1.90s | 1.45s


src/CMakeLists.txt
```diff
@@ -283,7 +284,6 @@ ncnn_add_layer(Quantize)
 ncnn_add_layer(Dequantize)
 ncnn_add_layer(Yolov3DetectionOutput)
 ncnn_add_layer(PSROIPooling)
-ncnn_add_layer(ROIAlign OFF)
```
nihui, line 286: Do not change the order
the generated layer index enum order breaks compatibility

src/layer/roialign.cpp
```diff
@@ -20,6 +20,117 @@ namespace ncnn {
...
+  int pos3;
+  int pos4;
+  T w1;
+  T w2;
+  T w3;
+  T w4;
+};
+
+template <typename T>
+void pre_calc_for_bilinear_interpolate(
```
nihui, line 38: The accelerated version shall be put into x86/arm/... subclass implementation
Keep src/layer/xyz.cpp simple and naive, and helpful for autotest

wkcn:
The result of naive ROIAlign is not consistent with that of x86 optimized one. I am trying to fix it.

wkcn:
The calculation of ROIAlign in ncnn is different from that in detectron2. The difference is mainly the calculation on the boundary and the fraction of the output value.
It is difficult to merge them.

nihui:
add new parameter to indicate the operator style
priorbox op uses the style param to distinguish mxnet and caffe

wkcn:
I add a new parameter `version` to decide the implementation of ROIAlign.
When `version = 0`, the original version of ROIAlign is invoked.
When `version = 1`, the detectron2 version of ROIAlign is invoked.

src/layer/roialign.cpp
wkcn: solved : )

nihui:
Thanks

PR #1823, opened 2020-06-08
fix cmake

wkcn:
Hi there,
When building ncnn, it outputs the following message:
```
CMake Error at src/CMakeLists.txt:413 (if):
  if given arguments:

    "DEFINED" "ANDROID_NDK_MAJOR" "AND" "GREATER" "20"

  Unknown arguments specified


-- Configuring incomplete, errors occurred!
```
I try to fix it.


nihui:
赞美大佬！

PR #1836, opened 2020-06-11
Added AVX implementation to cast to/from bfloat and float32

Timen:
Major changes:
1. added -mavx2 compiler option if NCNN_AVX2 is defined
2. added avx implementation of bfloat16 to float32 casting

AVX2 is required for this operation: _mm256_srli_epi32

@nihui I'm looking to start work towards better x86 optimizations, do you have some input on where to start? I was thinking to start by optimizing the different convolutions similar to the arm version but using only intrinsics, and adding an x86 version of the pooling layer. Currently it seems all convolutional layers use the same im2col sgemm implementation.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -399,7 +399,7 @@ elseif(NCNN_AVX2)
         target_compile_options(ncnn PRIVATE /arch:AVX2 /DAVX2)
     #Linux
     else()
-        target_compile_options(ncnn PRIVATE -mfma -mf16c)
+        target_compile_options(ncnn PRIVATE -mfma -mf16c -mavx2)
     endif()
 endif()
 
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -399,7 +399,7 @@ elseif(NCNN_AVX2)
         target_compile_options(ncnn PRIVATE /arch:AVX2 /DAVX2)
     #Linux
     else()
-        target_compile_options(ncnn PRIVATE -mfma -mf16c -mavx2)
+        target_compile_options(ncnn PRIVATE -mfma -mf16c)
     endif()
 endif()
 
--- a/src/layer/x86/avx_mathfun.h
+++ b/src/layer/x86/avx_mathfun.h
@@ -28,6 +28,8 @@
 
   (this is the zlib license)
 */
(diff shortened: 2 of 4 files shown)


Timen:
It seems the AVX2 version on the linux machines is failing. From the output of the CI the only thing I can think of is that the `-mavx2` compiler option is causing issues. Not sure what the exact problem is as on macos all tests are passing and so are the tests on windows. Is it possible the linux machine used for CI doesn't support AVX2?

Timen:
GPU version doesn't seem to support bfloat16 so I disabled the ctest for those. However the AVX2 CI fails, is a mystery to me. I'm trying to reproduce the build environment but I somewhat suspect the `-mavx2` build flag to cause issues on the specific CI machine. 

Do the linux and macos CI machines actually support avx2 vector extensions? The windows CI machine seems to not have any issues.

nihui:
according to docs, github CI machines should support AVX2

https://help.github.com/en/actions/reference/virtual-environments-for-github-hosted-runners#about-github-hosted-runners

https://docs.microsoft.com/en-us/azure/virtual-machines/dv2-dsv2-series#dsv2-series

nihui:
You are welcome to contribute better x86 optimization!
Be free to implement any optimized routines as you like, as long as it passes auto tests  :smile:

Timen:
Okay, yes it seems the CI machines do support avx2 and I was simply using an aligned load on unaligned memory. Ready for review!

BUG1989:
Hi @Timen,

I'm the the original author of  the x86 performance operators. My previous plan was this:

- ConvolutionDepthwise
- Pooling
- FC

more attractive is Int8 Implement:

- Convolution Int8
- ConvolutionDepthwise Int8
- Pooling Int8
- FC Int8

Timen:
@BUG1989 Hello! 

Cool, yes I'm planning something similar, however int8 is not currently on my list of things to do. I'm personally more interested using FP16 (not brain half) and element packing for best performance while staying mostly compatible with FP32 trained models. 

My list for x86 optimizations is as followed:
FC (starting simple)
Pooling
DWConv
LSTM

After that I need to decide whether or not I want to look into element packing or fp16 weight storage.

nihui:
Thanks !

PR #1839, opened 2020-06-12
AVX innerproduct and pooling 2x2 versions

Timen:
Major changes:
1. Innerproduct with AVX implementation
2. Pooling 2x2 AVX2 implementation
3. AVX activation functions + convenience functions
4. Fixed avx_mathfun compilation

Still in a bit of a rough shape and I'm sure the CI builds will fail. I'm also open for suggestions regarding my use of the AVX2 only operation `_mm256_permutevar8x32_epi32` but I'm not sure what alternative I could use to do a pair-wise maximum operation.


Timen:
Small AVX mathfun fix on windows to prevent float issues discussed in this PR:
https://github.com/pytorch/pytorch/pull/6192

Timen:
ready for review 👍

nihui:
I saw that you enabled c++17 build flags. What c++17 features do you want to use?
For maximizing portability, we don't force new c++ compiler features to build ncnn library.

Timen:
I removed the c++17 requirement, I think I added that to attempt to fix the MSVC build errors but it seems it wasn't necessary in the end.

src/layer/x86/innerproduct_x86.h
```diff
@@ -0,0 +1,45 @@
...
+
+    virtual int create_pipeline(const Option &opt);
+    virtual int destroy_pipeline(const Option &opt);
+
+    virtual int forward(const Mat &bottom_blob, Mat &top_blob,
+                        const Option &opt) const;
+
+protected:
+    int forward_f16s(const Mat &bottom_blob, Mat &top_blob,
+                     const Option &opt) const;
```
nihui, line 37: There is not a forward_f16s definition

Timen:
Simply changing the MALLOC_ALIGN value to 32 doesn't seem to fix the alignment issue. Besides it doesn't seem like unaligned vs aligned loads incur a performance penalty:

http://chrisdglover.blogspot.com/2014/05/the-performance-of-unaligned-simd-loads.html

I'm not 100% following the allocator @nihui so maybe I'm doing something wrong by only changing the MALLOC_ALIGN value to 32.

nihui:
> Simply changing the MALLOC_ALIGN value to 32 doesn't seem to fix the alignment issue. Besides it doesn't seem like unaligned vs aligned loads incur a performance penalty:
> 
> http://chrisdglover.blogspot.com/2014/05/the-performance-of-unaligned-simd-loads.html
> 
> I'm not 100% following the allocator @nihui so maybe I'm doing something wrong by only changing the MALLOC_ALIGN value to 32.

I don't think you could change all the unaligned access to aligned access, but only the ones in plain loops in cast relu etc.

Yeah, it is usually not an penalty to do unaligned access. There's nothing bad about changing malloc alignment to 256bit.

nihui:
By the way, you will probably aware that the 256bit alignment is useful as you can always use aligned access in pack8 storage mat  :smile:

Timen:
> By the way, you will probably aware that the 256bit alignment is useful as you can always use aligned access in pack8 storage mat 😄

That is true! I'm curious to see what the performance gains are for a mobilenet based network for pack 8 vs pack 4. Was the original pack4 implementation inspired by this paper: https://arxiv.org/abs/1809.10170

nihui:
> > By the way, you will probably aware that the 256bit alignment is useful as you can always use aligned access in pack8 storage mat smile
> 
> That is true! I'm curious to see what the performance gains are for a mobilenet based network for pack 8 vs non packed. Was the original pack4 implementation inspired by this paper: https://arxiv.org/abs/1809.10170 ?

no
I think it is a common practice to wrap the inner-most dimension into SIMD native size.

PR #1856, opened 2020-06-19
armv8.2 infrastructure

nihui:
Let's build armv8.2 optimized variants unconditionaly


Timen:
I’m not 100% familiar with this work but ehat I’m seeing are half precision arithmetic operations. Would that then support 8 lane wide vector instructions on arm? And is this maybe an opportunity to support 4 lane and 8 lane on both arm and x86? 

On x86 we could also do 4 lanes with sse instructions. Maybe one could even reuse the elempack 4 and elempack 8 code using stubs for the specific build. SSE/AVX/NEON/NEON(fp16)

nihui:
> I’m not 100% familiar with this work but ehat I’m seeing are half precision arithmetic operations. Would that then support 8 lane wide vector instructions on arm? And is this maybe an opportunity to support 4 lane and 8 lane on both arm and x86?
> 
> On x86 we could also do 4 lanes with sse instructions. Maybe one could even reuse the elempack 4 and elempack 8 code using stubs for the specific build. SSE/AVX/NEON/NEON(fp16)

Unlike x86, arm neon intrinsic is easily optimized negatively by the compiler. This is why inline assembly is used in arm optimization code. In addition, the neon instruction is not exactly the same as sse/avx, such as "vector multiply by lane". Using separate code implementations allows for more freedom to cook a well optimization of a particular architecture.

PR #1878, opened 2020-06-25
Fix windows build

zhiliu6:
resolve namespace problem when NCNN_AVX2 is enabled.


nihui:
Thanks for your contribution!

PR #1891, opened 2020-06-29
Fix AVX wino 3x3 and improve convolution test converage

Timen:
I was able to reproduce the 3x3 avx winograd error on the squeezenet example @nihui . Turned out that one of the tile processing loops (the one which only process one tile at a time to be precise) was incorrect. However this was never picked up by the ctest as it was not covering the tiles != 4 test cases.

In increasing the ctest coverage it also uncovered some other failing code paths, most notably the forwarddilation_x86 path. For now I have disabled that path, but it seems some kind of memory issue is at fault but I'm not 100% sure. 

One other problem it uncovered was that the fp16 3x3 winograd implementation for large inputs would deviate by more than 0.2f from the reference implementation. The reason this is happening is that while the reference weights are converted to and from fp16, it doesn't actually go through the same winograd kernel transformation before being used in the forward pass. The fp16 avx 3x3 winograd path actually applies a transformation on the kernel before converting it to fp16, and it seems that the values this transform outputs simply don't convert well to fp16. (Probably because the transformed kernel weights are laying further from zero which doesn't work well with the smaller exponent of fp16) 

I have disabled the forward dilation and fp16 winograd 3x3 kernel for now. I may revisit in a future PR.


nihui:
The numeric error caused by the fp16 precision lost will be enlarged after the weight transform and input/output transform.
That's why fp32 storage is used for transformed data in the arm bf16s winograd convolution. Thus, only the input transform read and output transform write benefits from memory bandwidth reducing.

src/layer/x86/convolution_x86.cpp
```diff
@@ -120,55 +119,56 @@ int Convolution_x86::create_pipeline(const Option& opt)
...
+    //         ncnn::Mat weights[1];
+    //         weights[0] = weight_data;
+
+    //         convolution_dilation1->load_model(ModelBinFromMatArray(weights));
+    //     }
+
+    //     convolution_dilation1->create_pipeline(opt);
+
+    //     return 0;
+    // }
```
nihui, line 171: prefer ```#if 0``` .... ```#endif``` macro for commenting large code blocks

tencent-adm:
[CLA assistant check](https://cla.opensource.tencent.com/Tencent/ncnn?pullRequest=1891) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you all sign our [Contributor License Agreement](https://cla.opensource.tencent.com/Tencent/ncnn?pullRequest=1891) before we can accept your contribution.<br/>**1** out of **2** committers have signed the CLA.<br/><br/>:white_check_mark: Timen<br/>:x: restyled-commits<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.opensource.tencent.com/check/Tencent/ncnn?pullRequest=1891) it.</sub>

src/layer/x86/convolution_x86.cpp
Timen: done!

Timen:
> The numeric error caused by the fp16 precision lost will be enlarged after the weight transform and input/output transform.
> That's why fp32 storage is used for transformed data in the arm bf16s winograd convolution. Thus, only the input transform read and output transform write benefits from memory bandwidth reducing.

Yeah I figured that, I was surprised by the large range of values (+/-10000) used in the transformed featuremap. Which is fine with fp32 however with fp16 this indeed causes issues.

Timen:
So it seems the ctest are now breaking here on the ARM 64 build:
`test_layer failed cpu
test_layer Convolution failed use_packing_layout=1 use_fp16_packed=1 use_fp16_storage=1 use_shader_pack8=1 use_bf16_storage=1 use_image_storage=1
test_convolution failed w=18 h=17 c=13 outch=4 kernel=3 dilation=1 stride=1 pad=1 bias=1 act=4 actparams=[-0.965801,0.043476]
`

I haven't really touched that code so it is probably a result of the increased coverage in the ctest. Maybe @nihui could have a look? It could be a a similar issue to what I had in the fp16 winograd 3x3 AVX kernel due to the usage of bf16 featuremaps (even though it doesn't use bf16 for weight storage).

nihui:
> So it seems the ctest are now breaking here on the ARM 64 build:
> `test_layer failed cpu test_layer Convolution failed use_packing_layout=1 use_fp16_packed=1 use_fp16_storage=1 use_shader_pack8=1 use_bf16_storage=1 use_image_storage=1 test_convolution failed w=18 h=17 c=13 outch=4 kernel=3 dilation=1 stride=1 pad=1 bias=1 act=4 actparams=[-0.965801,0.043476] `
> 
> I haven't really touched that code so it is probably a result of the increased coverage in the ctest. I can have a look soonish, even though i'm not 100% familiar with the arm code. It could be a a similar issue to what I had in the fp16 winograd 3x3 AVX kernel due to the usage of bf16 featuremaps (even though it doesn't use bf16 for weight storage).

fixed

PR #1897, opened 2020-06-30
Added AVX swish/lrn/batchnorm 

Timen:
Added the AVX version of swish lrn and batchnorm.

And increase convolution test coverage + unrolled some more loops.

### DIFF
--- a/tests/test_convolution.cpp
+++ b/tests/test_convolution.cpp
@@ -85,6 +85,7 @@ static int test_convolution_0()
                   || test_convolution(9, 7, 8, 4, kdsp[i][0], kdsp[i][1], kdsp[i][2], kdsp[i][3], 1)
                   || test_convolution(9, 7, 8, 13, kdsp[i][0], kdsp[i][1], kdsp[i][2], kdsp[i][3], 0)
                   || test_convolution(9, 7, 13, 8, kdsp[i][0], kdsp[i][1], kdsp[i][2], kdsp[i][3], 1)
+                  || test_convolution(9, 7, 4, 16, kdsp[i][0], kdsp[i][1], kdsp[i][2], kdsp[i][3], 0)
                   || test_convolution(9, 7, 16, 16, kdsp[i][0], kdsp[i][1], kdsp[i][2], kdsp[i][3], 0);
 
         if (ret != 0)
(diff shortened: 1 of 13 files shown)


tencent-adm:
[CLA assistant check](https://cla.opensource.tencent.com/Tencent/ncnn?pullRequest=1897) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you all sign our [Contributor License Agreement](https://cla.opensource.tencent.com/Tencent/ncnn?pullRequest=1897) before we can accept your contribution.<br/>**1** out of **2** committers have signed the CLA.<br/><br/>:white_check_mark: Timen<br/>:x: restyled-commits<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.opensource.tencent.com/check/Tencent/ncnn?pullRequest=1897) it.</sub>

tests/test_convolution.cpp
```diff
@@ -257,5 +259,5 @@ static int test_convolution_1()
 int main()
 {
     SRAND(7767517);
-    return test_convolution_0() || test_convolution_1() || test_convolution_2();
+    return test_convolution_2() || test_convolution_1() || test_convolution_0();
```
nihui, line 262: order reversed  :smile:

nihui:
looks good to me

tests/test_convolution.cpp
Timen: I know but otherwise it fails🤭

Timen: I suspect some alllocator issue, I will investigate more :) It was weird as it didn't fail on the ctest coverage I added but only a couple of subtests later. And it is fixed as I said by switching the order around 🤷‍♂️

Timen: Undid the re-order and actually fixed the issue :)

Timen:
I managed to track down the issue with the failing ctest to the sigmoid function producing nans. It seems that a specific set of randomly generated weight and input matrices would cause a very large exponent (>80) which the `exp()` function would return as a nan. 

The fix was to simply reduce the float range of the randomly generated matrices from -2/2 to -1.5/1.5.

nihui:

PR #1909, opened 2020-07-03
remove x86 reshape layers

Timen:
I ran into a regression using the reshape avx layers. I won't have time to look into it in the near future so I thought it would be best to remove them for now till I or someone else has a proper look at it and adds tests to cover the regression. It seemed something to do with tensorflow style reshape on a packed blob.


nihui:
For the tensorflow style reshape, do you mean reshape with param 3=1 ？

Timen:
Correct, and I also see that this value is not actually tested outside the default value in the ctest. I'm not able to contribute in the foreseeable future but would be good is someone could add a test case covering the permute value being set to 1 instead of zero.

nihui:
reshape with permute implemented and tests included

Timen:
@nihui thanks for picking this up so fast!

PR #1916, opened 2020-07-05
update readme: metion yolact, retinaface, etc

zchrissirhcz:
According to the source files in `example` directory,  networks such as yolact, retinaface are added to readme

Also link license to LICENSE.txt file instead of pure text.


zchrissirhcz:
The second commit is for `doc` folders update:  add `c/cpp/bash/python/cmake` to code region/blocks, highlighting code lines in documention files.

nihui:
谢谢！

PR #1943, opened 2020-07-19
avx2 infrastructure

nihui:


Timen:
One thing that may be a good idea to add here is add a build target for AVX, and a separate one for AVX2. As older cpus (like sandy bridge) don't support the AVX2 features. The major features missing in AVX are Fused multiply accumulate, and float16 float32 conversion operations.

Simplest thing would be to replace the fmadd operation with one using a multiply and addition when building for AVX and disable the FP16 operations.

nihui:
> One thing that may be a good idea to add here is add a build target for AVX, and a separate one for AVX2. As older cpus (like sandy bridge) don't support the AVX2 features. The major features missing in AVX are Fused multiply accumulate, and float16 float32 conversion operations.
> 
> Simplest thing would be to replace the fmadd operation with one using a multiply and addition when building for AVX and disable the FP16 operations.

Yes, only dynamic dispatch of avx2 path is supported for the time being. cpu without avx2 uses the x86-64 path.
If needed, the avx path can be added in the future.

nihui:
> One thing that may be a good idea to add here is add a build target for AVX, and a separate one for AVX2. As older cpus (like sandy bridge) don't support the AVX2 features. The major features missing in AVX are Fused multiply accumulate, and float16 float32 conversion operations.
> 
> Simplest thing would be to replace the fmadd operation with one using a multiply and addition when building for AVX and disable the FP16 operations.

ping Timen

To thank you for your avx optimization contribution to ncnn, I have contacted TOS to issue you a Contributor Certificate. However, I need your contact information to mail the item. Please reach me via nihuini at tencent dot.com.

PR #1957, opened 2020-07-22
Add a option to allow manually block AndroidAsset related codes, for complie in Termux

tanyiok1234:
Because  `__ANDROID_API__ >= 9` is usually true in termux, `AndroidAsset` will introduces complie errors.

So I simply added a option "NOT_TERMUX" to control whether block these part or not manually.


tencent-adm:
[CLA assistant check](https://cla.opensource.tencent.com/Tencent/ncnn?pullRequest=1957) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.opensource.tencent.com/Tencent/ncnn?pullRequest=1957) before we can accept your contribution.<br/><hr/>**tanyi** seems not to be a GitHub user. You need a GitHub account to be able to sign the CLA. If you have already a GitHub account, please [add the email address used for this commit to your account](https://help.github.com/articles/why-are-my-commits-linked-to-the-wrong-user/#commits-are-not-linked-to-any-user).<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.opensource.tencent.com/check/Tencent/ncnn?pullRequest=1957) it.</sub>

nihui:
The NOT_TERMUX option seems to be a quite hacky stuff.

How about setting the default compile deifinitions before calling cmake ?

```
export CFLAGS=-D__ANDROID_API__=0
export CXXFLAGS=-D__ANDROID_API__=0
```

tanyiok1234:
> The NOT_TERMUX option seems to be a quite hacky stuff.
> 
> How about setting the default compile deifinitions before calling cmake ?
> 
> ```
> export CFLAGS=-D__ANDROID_API__=0 
> export CXXFLAGS=-D__ANDROID_API__=0 
> ```

The reason why I choose this hacky way is because I noticed that it is not the only place where this condition applies (like gpu & vulkan related code)....

You know, it still needs to link the lib inside system. Because rest of the code looks fine, this method looks less harmful. (Also because I am not sure what's on there)

I know it looks ... well you know what I want to say. That's exactly why I hesitated about whether submit a pull request or not for months. Or you can just leave it open, so someone who have the same issues can get a hacky way, as this won't be the only change required for a Termux complie.

Btw, I tried the compile definitions you proposed just now, which report new errors.
I am pretty sure the complie flag is the only difference. 
So it is not a easy solution, at least for now.

nihui:
So what was the original compilation error in the Android Asset Manager API?
If Termux automatically defines __ANDROID_API__, these Android API should be allowed to be used.

tanyiok1234:
> So what was the original compilation error in the Android Asset Manager API?
> If Termux automatically defines ```__ANDROID_API__```, these Android API should be allowed to be used.

Yes, it is allowed, but needs to link lots useless additional  libraries. It is the libandorid plus lots of libs related to it (Sorry I forget the details).

Without linking to the libandroid:



Linking to libandroid but not others: 


Plus, it is something changes with android vesion (different in Oreo and Pie). 

Termux provides a linux like file I/O mechanisn, I think not one would use the AndriodAsset in a termux program.

nihui:
okay, for this particular usecase, it makes sense to disable all platform-specific api uses

* add new cmake option ```NCNN_PLATFORM_API``` defaulting to ON
* add new macro definition to platform.h.in
* guard all platform-specific code with NCNN_PLATFORM_API
such as ```#if __ANDROID_API__ >= 9``` to ```#if NCNN_PLATFORM_API && __ANDROID_API__ >= 9```

nihui:
move to https://github.com/Tencent/ncnn/pull/2502

PR #2048, opened 2020-08-27
Compiler flag changes and other minor things

Artoria2e5:
* CMakeList:
  * Use fast fp on MSVC, just like we do for GNU.
  * Since fast fp does not interfere with debug, enable it always.
  * Set RPI3 flags correctly (not native, but the actual model)
  * Uncomment the enable IPO thing as we can now test for support.
  * Prepare for `#pragma omp simd`. Our current compiler options does enable auto-vectorization on release, however.
* mat.h, hedley.h:
  * Mechanism for labelling alignment. We already have MALLOC_ALIGN, but the compiler needs to know for better vectorization.
  * Turns out we cannot label alignment, becuase the rows can break them. Uh-oh.

### DIFF
--- a/src/allocator.h
+++ b/src/allocator.h
@@ -45,6 +45,9 @@ namespace ncnn {
 #define MALLOC_ALIGN 16
 #endif
 
+// For marking allocated things as aligned. Will evaluate twice on GCC!
+#define NCNN_ASSUME_ALIGNED(ptr) HEDLEY_ASSUME_ALIGNED(ptr, MALLOC_ALIGN)
+
 // Aligns a pointer to the specified number of bytes
 // ptr Aligned pointer
 // n Alignment size that must be a power of two
@@ -63,6 +66,7 @@ static inline size_t alignSize(size_t sz, int n)
     return (sz + n - 1) & -n;
 }
 
+// Aligned malloc.
 static inline void* fastMalloc(size_t size)
 {
 #if _MSC_VER
--- a/src/mat.h
+++ b/src/mat.h
@@ -668,6 +668,8 @@ inline Mat::Mat(const Mat& m)
         NCNN_XADD(refcount, 1);
 }
 
+// With these constructors I cannot prove that data is aligned to MALLOC_ALIGN...
+// Unless, of course, we yell at users to guarantee these.
 inline Mat::Mat(int _w, void* _data, size_t _elemsize, Allocator* _allocator)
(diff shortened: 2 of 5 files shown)


tencent-adm:
[CLA assistant check](https://cla.opensource.tencent.com/Tencent/ncnn?pullRequest=2048) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.opensource.tencent.com/Tencent/ncnn?pullRequest=2048) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.opensource.tencent.com/check/Tencent/ncnn?pullRequest=2048) it.</sub>

nihui:
please fix the ci errors

src/allocator.h
```diff
@@ -45,6 +46,9 @@ namespace ncnn {
 #define MALLOC_ALIGN 16
 #endif
 
+// For marking allocated things as aligned. Will evaluate twice on GCC!
+#define NCNN_ASSUME_ALIGNED(ptr) HEDLEY_ASSUME_ALIGNED(ptr, MALLOC_ALIGN)
+
```
nihui, line 51: unused macro

src/mat.h
```diff
@@ -39,6 +39,14 @@
...
 #endif // NCNN_PIXEL
 
+#if !defined(HEDLEY_MSVC_VERSION) || HEDLEY_MSVC_VERSION_CHECK(19, 20, 0)
+#define NCNN_OMP_SIMD            HEDLEY_PRAGMA(omp simd)
+#define NCNN_OMP_SIMD_EX(Extra)  HEDLEY_PRAGMA(omp simd Extra)
+#else
+#define NCNN_OMP_SIMD
+#define NCNN_OMP_SIMD_EX(Extra)
+#endif
+
```
nihui, line 49: How is the actual speedup benefit of omp simd clause ?
We use explicit simd intrinsic and inline assembly for most critical code paths, I don't think the compiler could do better.

nihui (changes requested):

src/mat.h
Artoria2e5: There are a lot of generic code still present in here, and for simple cases auto-vectorization is less error-prone. This also speeds up debug-mode since it's previously not vectorized. MSVC also specifically tells you why something could not be vectorized when you add this.

src/allocator.h
Artoria2e5: I tried to use it, but see my comments in Mat.h. I could not really proved that the derived rows or columns are aligned. I would really with to be able to use it.

PR #2071, opened 2020-09-04
NCNN_BUILD_TESTS set default to OFF

caishanli:
most of us don't need to build tests


caishanli:
so how to enable test in all github action ci

nihui:
> so how to enable test in all github action ci

add ```-DNCNN_BUILD_TESTS=ON``` cmake option to all build phase
ci configuration files are in ncnn/.github/workflows/xxx.yml

PR #2076, opened 2020-09-07
set NCNN_BUILD_TESTS default off 

caishanli:
set cmake NCNN_BUILD_TESTS default off and add -DNCNN_BUILD_TESTS=ON to ci


nihui:

PR #1973, opened 2020-07-27
Fix more than 64 cpu

tpoisonooo:
#### 主题
尝试修复 CPU 个数大于 64 时设置亲和力失败的问题。

a. 如果是 linux/unix 流派系统，引用系统库`sched.h`的`cpu_set_t`;
b. 如果是 windows 系统，自己糙一个 `struct cpu_set_t`。装作 64 核 windows 跑 ncnn 的场景不存在...

#### 测试
跑了 benchmark 没崩
图片

PR #2082, opened 2020-09-08
Fix cpu limit

tpoisonooo:
fix  more than 64-cpu
关联 pr  https://github.com/Tencent/ncnn/pull/1973


tpoisonooo:
@nihui  这是个新的，求 reopen 一下...

PR #2086, opened 2020-09-09
FAQ-ncnn-vulkan.md: fix outdated document

vowstar:
Nihui yes!

change document about how to enable ncnn vulkan capablity

Signed-off-by: Huang Rui <vowstar@gmail.com>


nihui:
Thanks !

PR #2104, opened 2020-09-12
ci ndk-r16b

nihui:


nihui:
depends on https://github.com/KhronosGroup/glslang/pull/2393

nihui:
depends on https://github.com/KhronosGroup/glslang/pull/2395

PR #2189, opened 2020-10-10
Fix protobuff version error with onnx2ncnn

deepage:
Fix protobuff version error with onnx2ncnn in issues[#2187 ](https://github.com/Tencent/ncnn/issues/2187)


nihui:
Thanks

PR #2209, opened 2020-10-13
fix darknet2ncnn.cpp warnings

zchrissirhcz:
fix two warnings:
1.  determine if fp!=NULL before fclose
2.  wrapping (size_t) for int\*int

### DIFF
--- a/tools/darknet/darknet2ncnn.cpp
+++ b/tools/darknet/darknet2ncnn.cpp
@@ -828,13 +828,14 @@ void load_weights(const char* filename, std::deque<Section*>& dnet)
             }
 
             if (s->layer_type == "Convolution")
-                read_to(s->weights, s->c * s->filters * s->size * s->size, fp);
+                read_to(s->weights, (size_t)(s->c) * s->filters * s->size * s->size, fp);
             else if (s->layer_type == "ConvolutionDepthWise")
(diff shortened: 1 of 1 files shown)


tools/darknet/darknet2ncnn.cpp
```diff
@@ -828,13 +828,14 @@ void load_weights(const char* filename, std::deque<Section*>& dnet)
...
             if (s->layer_type == "Convolution")
-                read_to(s->weights, s->c * s->filters * s->size * s->size, fp);
+                read_to(s->weights, (size_t)(s->c) * s->filters * s->size * s->size, fp);
             else if (s->layer_type == "ConvolutionDepthWise")
                 read_to(s->weights, s->c * s->filters * s->size * s->size / s->groups, fp);
         }
     }
 
-    fclose(fp);
+    if (fp != NULL)
```
proydakov, line 837: I think this is an extra check. We have condition with NULL higher:

    FILE* fp = fopen(filename, "rb");
    if (fp == NULL)
        file_error(filename); // exit

zchrissirhcz: @proydakov Agree. I did this modification since Visual Studio 2019 give "warning hint" if not check. Trying to find reasonable explanation for this hint.

PR #2211, opened 2020-10-13
fix roialign_x86.cpp integer multiply may overflow

zchrissirhcz:


ncnnnnn:
试一下按照 Multiplication result may overflow 'int' before it is converted to 'size_type'.
结果直接pr了。。。真是泪奔。。。。。。

nihui (approved):

nihui:
Thanks !

PR #2213, opened 2020-10-13
fix packing.cpp integer multiply may overflow

zchrissirhcz:


nihui (commented):
please check https://github.com/Tencent/ncnn/security/code-scanning/40?query=ref%3Arefs%2Fpull%2F2213%2Fhead

nihui (approved):

nihui:
Thanks !

PR #2236, opened 2020-10-20
add interp param align_corner

maxfy1992:
- add interp op type type 4=bilinear+align corner=true (pytorch Upsampling Op align_corners=True )
- add onnx2ncnn for model convert


nihui:
add a new param member ```int align_corner``` in Interp
modify test_interp.cpp to include this new type
besides, if it does not have a vulkan implementation for the time being, add ```if (type == 4) support_vulkan = false``` in ```Interp_vulkan::create_pipeline()```

maxfy1992:
got it, I will fix it as soon as I can

maxfy1992:
how to re-run ci checks?

maxfy1992:
add if (type == 4) support_vulkan = false in Interp_vulkan::create_pipeline() seems not work for test_interp.cpp.

nihui:

PR #2253, opened 2020-10-25
Fix build instructions

pH5:
Add the missing -D option character for the NCNN_OPENCV and NCNN_BUILD_EXAMPLES options to the example cmake command line.


nihui (changes requested):
```-DNCNN_OPENCV=ON``` option is obsoleted, drop it

nihui (approved):

nihui:
Thanks !

PR #2254, opened 2020-10-25
Silence unused parameter warning in opencv resize function

pH5:
Silence the unused parameter warning:
```
../src/opencv.cpp: In function ‘void cv::resize(const cv::Mat&, cv::Mat&, const cv::Size&, float, float, int)’:
../src/opencv.cpp:83:81: warning: unused parameter ‘flags’ [-Wunused-parameter]
   83 | void resize(const Mat& src, Mat& dst, const Size& size, float sw, float sh, int flags)
      |                                                                             ~~~~^~~~~
```


nihui:

PR #2255, opened 2020-10-25
Fix maybe uninitialized warnings in Vulkan reshape functions

pH5:
Fix these warnings (they do look like false positives):
```
../src/layer/vulkan/reshape_vulkan.cpp: In member function ‘virtual int ncnn::Reshape_vulkan::forward(const ncnn::VkImageMat&, ncnn::VkImageMat&, ncnn::VkCompute&, const ncnn::Option&) const’:
../src/layer/vulkan/reshape_vulkan.cpp:925:13: warning: ‘out_elempack’ may be used uninitialized in this function [-Wmaybe-uninitialized]
  925 |             if (out_elempack == 1) out_elemsize = 4u;
      |             ^~
```


nihui:
Thanks !

PR #2272, opened 2020-11-02
fix: fix gnueabihf cmake toolchain neon flags

zchrissirhcz:
Test platform: ZYNQ 7020, there is vfpv3 but isn't vfpv4
(Reported in QQ Group, 2020-11-02)


nihui:
Thanks !

PR #2273, opened 2020-11-02
try merge static inline int NCNN_XADD

ncnnnnn:


nihui:
move to https://github.com/Tencent/ncnn/pull/2279

PR #2284, opened 2020-11-05
maybe benchncnn -DBL_MAX2DBL_MIN?

ncnnnnn:


nihui:
DBL_MIN is the smallest **positive** normal double

PR #2288, opened 2020-11-05
Extract on CPU without pack/fp16fp32

MarsTechHAN:


nihui (changes requested):
```int _unchange```
change the parameter identifier to a more meaningful one

MarsTechHAN:
> `int _unchange`
> change the parameter identifier to a more meaningful one

@nihui How about `int extract_without_modify`?

MarsTechHAN (commented):
Change to type and add comments in net.h

nihui:
> Change to type and add comments in net.h

also apply your changes to net.cpp

nihui (approved):

PR #2290, opened 2020-11-05
fix can't build on vs2013

corleonechensiyu:
add x86 AVX fp16


nihui:
please fix auto test

tpoisonooo:
bug-fix and `AVX fp16` feature should be different PR  +_+

PR #2311, opened 2020-11-12
Added unittest for cpu module. Improved code coverage.

proydakov:
Hi, NCNN Team.

As I see the src/cpu.cpp has the least code coverage: https://codecov.io/gh/Tencent/ncnn/src/master/src/cpu.cpp#
I added several unittest for this subsystem.

Could you verify and accept my changes, pls?

Best regards, Proydakov Evgeny.


nihui:

PR #2314, opened 2020-11-13
Update README.md

LosReturn:
Update README.md for Netron.app


nihui:

PR #2333, opened 2020-11-20
cmake: change the scope of -fno-exceptions and -fno-rtti

chosungmann:
ToT declares -fno-exceptions and -fno-rtti compiler options as a PUBLIC item,
but the compiler options in the public scope affects the compilation process of
the other projects doing add_executable(ncnn) or add_library(ncnn). This patch
eliminates the potential risks by changing the scope of the options above.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -269,11 +269,11 @@ else()
 endif()
 
 if(NCNN_DISABLE_RTTI)
-    target_compile_options(ncnn PUBLIC -fno-rtti)
+    target_compile_options(ncnn PRIVATE -fno-rtti)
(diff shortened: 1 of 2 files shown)


chosungmann:
Feel free to close this, if you think the issue was intended!

nihui:
If ncnn is built without RTTI or exceptions, then executable and shared library linking ncnn has to be built without RTTI or exception. Otherwise, you will probably get compile error like "undefined reference to type_info xxxx" or something like, as you can find them in the failing ci.

We disable RTTI and exceptions in default build for minimal binary size

nihui:
Thanks anyway

PR #2335, opened 2020-11-20
support Apple M1 Silicon

nihui:
- [x] recognize apple m1
~~- [ ] cpu affinity on macos/ios~~

### DIFF
--- a/src/cpu.cpp
+++ b/src/cpu.cpp
@@ -45,12 +45,12 @@
 
 #if __APPLE__
 #include <mach/mach.h>
-#include <mach/thread_act.h>
-#include "TargetConditionals.h"
-#if TARGET_OS_IPHONE
 #include <mach/machine.h>
+#include <mach/thread_act.h>
 #include <sys/sysctl.h>
 #include <sys/types.h>
+#include "TargetConditionals.h"
+#if TARGET_OS_IPHONE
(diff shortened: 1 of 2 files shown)


src/cpu.cpp
```diff
@@ -487,6 +530,40 @@ static int set_sched_affinity(const CpuSet& thread_affinity_mask)
 }
 #endif // defined __ANDROID__ || defined __linux__
 
+#if __APPLE__
+static int set_sched_affinity(const CpuSet& thread_affinity_mask)
+{
+    // https://developer.apple.com/library/archive/releasenotes/Performance/RN-AffinityAPI/index.html
```
AmeAkio, line 536: This is a quite outdated document. Apple will not allow developers to set CPU affinity. In OS X 10.5 it worked, later it became a suggestion to OS X, then in 10.10 or so (as well in later ones), macOS will ignore any affinity settings.

nihui: It returns KERN_NOT_SUPPORTED
QAQ

AmeAkio: It seems that we have no choice but trusts the thread scheduler in macOS. (I also tried `processor_exit` to shutdown some cores, but that only works on the Intel ones.) I'm afraid that there's no way for us to do this...

PR #2337, opened 2020-11-20
update operators.md,add argmax and bias

xingxingRealzyx:


nihui:
thanks for your contribution

PR #2359, opened 2020-11-26
update operators.md, add binaryop

xingxingRealzyx:


nihui:
please add binray operation type doc

nihui:
Thanks !

PR #2361, opened 2020-11-27
try onnx2ncnn add softplus

ncnnnnn:
onnx/backend/test/data/node/test_softplus_example/model.onnx
x = np.array([-1, 0, 1]).astype(np.float32)
ncnn 0.313262 0.693147 1.313262 
onnx 0.31326166, 0.69314718, 1.31326163


nihui:
Thanks !

PR #2365, opened 2020-11-27
Support negative axis in concat, slice and softmax

Ca0L:


nihui (approved):

nihui:

PR #2374, opened 2020-11-30
change abs() to fabs() for float args on ARM

zchrissirhcz:
For compatiblity for old compilers such as ndk-r16b.
`fabs()` keeps result stable.

ref: https://github.com/Tencent/ncnn/issues/2372


nihui:
Thanks !

PR #2378, opened 2020-12-02
optimize left over x86 convolution

zhiliu6:
Non-optimized convolution will be 5-50% faster~


tpoisonooo:
`optimized convolution will be 5-50% faster`

nihui:
Thanks !

PR #2383, opened 2020-12-04
add python

caishanli:
add python to ncnn


CMakeLists.txt
```diff
@@ -175,3 +176,6 @@ if(NCNN_BUILD_TESTS)
     enable_testing()
     add_subdirectory(tests)
 endif()
+if(NCNN_BUILD_PYTHON)
+	add_subdirectory(python)
```
nihui, line 180: space instead of tab

CMakeLists.txt
```diff
@@ -53,6 +53,7 @@ option(NCNN_DISABLE_PIC "disable position-independent code" OFF)
 option(NCNN_BUILD_TESTS "build tests" OFF)
 option(NCNN_COVERAGE "build for coverage" OFF)
 option(NCNN_BUILD_BENCHMARK "build benchmark" ON)
+option(NCNN_BUILD_PYTHON "build python api" OFF)
```
nihui, line 56: what about NCNN_PYTHON

python/ncnn/utils/download.py
```diff
@@ -0,0 +1,114 @@
+"""Download files with progress bar."""
+# Tencent is pleased to support the open source community by making ncnn available.
```
nihui, line 2: license header first

PR #2389, opened 2020-12-05
don't compile tools and examples when crosscompile

zchrissirhcz:
When cross-compiling, such as arm-linux-gnueabihf,
x86 OpenCV may be wrongly found by tools & examples targets.
Turn tools and examples off when cross compiling.


nihui (approved):

nihui:
Thanks !

Issue #2432, opened 2020-12-11
iOS 集成ncnn bitcode的framework，上传到app store报错


zhfwch:
报错详细信息：
> ITMS-90562: Invalid Bundle - The app cannot be processed because options not allowed to be embedded in bitcode are detected in the submission. It is likely that you are not building the app with the toolchain provided in Xcode. Rebuild your entire app with the latest GM Xcode and submit the app again.

依照网上解决方案[bitcode ITMS-90562: Invalid Bundle](https://easeapi.com/blog/blog/104-invalid-bundle.html)无效。暂时用非bitcode打包上传到app store。

烦请后续版本解决一下。谢谢！

nihui:
尝试最新的release，是用xcode编译的

PR #2440, opened 2020-12-14
Ci release

nihui:
- [x] embed version in zip filename
- [x] ~~combined glslang for ubuntu/windows/android and cmake config~~
- [x] teach cmake config to find glslang in lib64
- [x] build option and runtime function for ncnn version
- [x] ~~drop NCNN_RUNTIME_CPU NCNN_SSE2 NCNN_AVX2 NCNN_ARM82 macro in platform.h (unified header)~~
- [x] test openmp avaliablity on macos and ios
- [x] test bitcode working #2432


nihui:
```
# android and linux
libtool -static libx.a liby.a libz.a -o libxyz.a
```

```
# windows
lib /out:xyz.lib x.lib y.lib z.lib
```

PR #2456, opened 2020-12-15
update AMD 3970X + RTX8000 benchmark

BUG1989:
- update the CPU benchmark with AMD 3970X
- update the GPU benchmark with NVIDIA QUADRO RTX 8000


nihui:
Thanks !

PR #2474, opened 2020-12-18
add destroy function to custom layer

caishanli:
store custom layer name string to layer_registry_entry

### DIFF
--- a/src/layer.h
+++ b/src/layer.h
@@ -144,15 +144,17 @@ public:
 
 // layer factory function
 typedef Layer* (*layer_creator_func)();
+typedef void (*layer_destroyer_func)(Layer*);
 
 struct layer_registry_entry
 {
 #if NCNN_STRING
     // layer type name
-    const char* name;
+    std::string name;
 #endif // NCNN_STRING
     // layer factory entry
     layer_creator_func creator;
+    layer_destroyer_func destroyer;
 };
 
 #if NCNN_STRING
--- a/src/layer.h
+++ b/src/layer.h
@@ -127,6 +127,8 @@ public:
 public:
     // layer type index
     int typeindex;
+    // custom layer creator index
+    int custom_layer_creator_index;
 #if NCNN_STRING
     // layer type name
     std::string type;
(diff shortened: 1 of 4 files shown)


src/layer.h
```diff
@@ -144,15 +144,17 @@ class Layer
...
 // layer factory function
 typedef Layer* (*layer_creator_func)();
+typedef void (*layer_destroyer_func)(Layer*);
 
 struct layer_registry_entry
 {
 #if NCNN_STRING
     // layer type name
-    const char* name;
+    std::string name;
```
nihui, line 153: what is the rationale of using non-POD type here

caishanli: in c/c++, string compiled to .text section, and we can use it(all the same string have same address) in program anywhere.
but in python, if a string object have no reference(no name), it's reference count is 0 after pass to c/c++, and recycled by gc, for example:
`net.register_custom_layer("YoloV5Focus", YoloV5Focus_layer_creator)`
during this call, net can see the string "YoloV5Focus", but after this call, "YoloV5Focus" is recycled, and net.load_param can't find this registed layer by name.

so anyother solutions?

src/layer.h
```diff
@@ -170,6 +174,15 @@ Layer* create_layer(int index);
         return new name;                  \
     }
 
+#define DEFINE_LAYER_DESTROYER(name)      \
+    void name##_layer_destroyer(::ncnn::Layer* layer) \
+    {                                     \
+        if (layer)                        \
+        {                                 \
+            delete layer;                 \
+        }                                 \
```
nihui, line 183: This condition is not necessary
It is fine to delete NULL

src/layer.h
```diff
@@ -127,6 +127,8 @@ class Layer
 public:
     // layer type index
     int typeindex;
+    // custom layer creator index
+    int custom_layer_creator_index;
```
nihui, line 131: what about using (ncnn::LayerType::CustomBit | index) for typeindex to indicate the custom index ?

src/net.cpp
```diff
@@ -1104,7 +1113,9 @@ Layer* Net::create_custom_layer(int index)
     if (!layer_creator)
         return 0;
 
-    return layer_creator();
+    Layer* layer = layer_creator();
+    layer->custom_layer_creator_index = index;
+    return layer;
```
nihui, line 1118: layer->typeindex = ncnn::LayerType::CustomBit | index;

PR #2478, opened 2020-12-18
Support Allwinner V831

sunnycase:
- Add v831 support to build script
- Use `getpid` instead of `gettid` due to v831 toolchain
- Add v831 toolchain file


nihui:
我来舔你啦.jpg

PR #2481, opened 2020-12-21
add custom layer destroyer

caishanli:


nihui:

PR #2494, opened 2020-12-23
fix src/net.cpp missing-field-initializers warning

zchrissirhcz:
Hi Nihui

On my local MacOSX machine, cross compiling for android arm64 platform, ndk-r21b, gives this warning:
```
[1/3] Building CXX object src/CMakeFiles/ncnn.dir/net.cpp.o
../../src/net.cpp:99:51: warning: missing field 'destroyer' initializer [-Wmissing-field-initializers]
        struct layer_registry_entry dummy = {"", 0};
```

This PR try to remove that warning.


nihui:
Thanks !

Issue #2523, opened 2020-12-29
build shared library


nihui:
- [x] cmake option for building shared library
- [x] cmake portable export macro
- [x] semantic version 1.0.20201228
- [x] build windows dll
- [x] build linux so
- [x] build macos dylib
- [x] build android so
- [x] ~~build ios dylib~~
- [x] Allocator ABI (pimpl + public data structure)
- [x] benchmark.h C ABI
- [x] Blob ABI stable (public data structure)
- [x] cpu.h C ABI
- [x] Command ABI (pimpl)
- [x] DataReader ABI (pimpl)
- [x] gpu.h C ABI (public data structure)
- [x] Layer ABI stable (reserved supports)
- [x] Mat ABI stable (public data structure)
- [x] ModelBin ABI (pimpl)
- [x] Net ABI (pimpl)
- [x] Option ABI (reserved uses)
- [x] ParamDict ABI (pimpl)
- [x] Pipeline ABI (public data structure)
- [x] PipelineCache ABI (pimpl)
- [x] platform.h (public data structure)
- [x] simpleocv.h (public data structure)
- [x] simpleomp.h C ABI
- [x] simplestl.h C ABI
- [x] glslang symbol visibility

PR #2525, opened 2020-12-30
build shared library

nihui:
see https://github.com/Tencent/ncnn/issues/2523


nihui:
- [ ] fix STL export warning
- [ ] more reserved for gpu info
- [ ] pipeline pimple

PR #2541, opened 2021-01-02
Fix python build

zchrissirhcz:
Hi, @caishanli  && @nihui 

When I try to compile ncnn with python (i.e. pyncnn), it failed to link.

My platform: 
- MacOSX, pybind11 from https://gitee.com/mirrors/pybind11
- Python 3.7.6

I just temporarily commented out two lines of code
```c++
    //.def_property("blobs", &Net::mutable_blobs)
    //.def_property("layers", &Net::mutable_layers);
```
which, I think, should be fixed later ( but I don't know how to fix).

Also updated the documentation, `DNCNN_BUILD_PYTHON` => `DNCNN_PYTHON` according to CMakeLists.txt

### DIFF
--- a/python/README.md
+++ b/python/README.md
@@ -23,7 +23,7 @@ git submodule init && git submodule update
 ```bash
 mkdir build
 cd build
-cmake -DNCNN_BUILD_PYTHON=ON ..
+cmake -DNCNN_PYTHON=ON ..
 make
 ```
 
--- a/python/src/main.cpp
+++ b/python/src/main.cpp
@@ -622,9 +622,9 @@ PYBIND11_MODULE(ncnn, m)
 #endif // NCNN_STDIO
 
     .def("clear", &Net::clear)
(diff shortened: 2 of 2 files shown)


zchrissirhcz:
I guess for the `blobs` and `layers` variable in Python API, it might be using something like:
```c++
.def_property("blobs", &Pet::getBlobs, &Pet::setBlobs)
```
instead of 
```c++
.def_property("blobs", &Net::blobs)
```
or
```c++
.def_property("blobs", &Net::mutable_blobs)
```

(ref: https://pybind11.readthedocs.io/en/stable/classes.html)

caishanli:
this is due to that Net::blobs and Net::layers changed from member variable to member function, fixed in [pr2550](https://github.com/Tencent/ncnn/pull/2550)

zchrissirhcz:
Ok, close due to fixed in metioned PR.

PR #2546, opened 2021-01-04
support PyTorch AdaptiveAvgPool2d and AdaptiveMaxPool2d

GuoxiaWang:
support pytorch adaptive pool
    support onnx2ncnn adaptive pool convert
    support ncnnoptimize adaptive pool param write


src/layer/pooling.cpp
```diff
@@ -99,6 +102,83 @@ int Pooling::forward(const Mat& bottom_blob, Mat& top_blob, const Option& opt) c
...
+                        for (int ih = ih0; ih < ih1; ih++)
+                        {
+                            for (int iw = iw0; iw < iw1; iw++)
+                            {
+                                sum += inptr[ih * w + iw];
+                            }
+                        }
+
+                        outptr[j] = sum / hk / wk;
+                    }
```
nihui, line 172: What about the interaction with the avgpool_count_include_pad parameter ?

tools/onnx/onnx2ncnn.cpp
```diff
@@ -3157,6 +3161,22 @@ int main(int argc, char** argv)
...
+                pool = 1;
+            }
+            int adaptive_pooling = 1;
+            const onnx::TensorProto& out_shape_tp = weights[node.input(1)];
+            std::vector<int> out_shape = get_node_attr_from_input_ai(out_shape_tp);
+
+            fprintf(pp, " 0=%d", pool);
+            fprintf(pp, " 7=%d", adaptive_pooling);
+            fprintf(pp, " 17=%d", out_shape[0]);
+            fprintf(pp, " 18=%d", out_shape[1]);
```
nihui, line 3178: Is the out_shape hw order or wh order ?

src/layer/pooling.cpp
```diff
@@ -99,6 +102,83 @@ int Pooling::forward(const Mat& bottom_blob, Mat& top_blob, const Option& opt) c
         return 0;
     }
 
+    if (adaptive_pooling)
+    {
+        top_blob.create(out_w, out_h, channels, elemsize, opt.blob_allocator);
+        if (top_blob.empty())
+            return -100;
```
nihui, line 109: Is it possible that explicit padding (pad_left pad_right ...) could be applied before pooling section ?

nihui (changes requested):
Please add testcase for the new adaptive pooling and output shape into tests/test_pooling.cpp

I see no vulkan implementation for these cases, claim support_vulkan = false in pooling_vulkan create_pipeline()

src/layer/pooling.cpp
GuoxiaWang: According PyTorch implementation, it do not need padding params. [AdaptiveAvgPoolKernel](https://github.com/pytorch/pytorch/blob/690eaf9c43f3f71051d90e9f47f25b014ce84681/aten/src/ATen/native/cpu/AdaptiveAvgPoolKernel.cpp#L33)

src/layer/pooling.cpp
GuoxiaWang: Because there are no padding params, so avgpool_count_include_pad is not be used.

tools/onnx/onnx2ncnn.cpp
GuoxiaWang: The out_shape is  NCHW, so it is hw order.

GuoxiaWang (commented):
> Please add testcase for the new adaptive pooling and output shape into tests/test_pooling.cpp
> 
> I see no vulkan implementation for these cases, claim support_vulkan = false in pooling_vulkan create_pipeline()

Ok, I will add a test case, wait a moment.

src/layer/pooling.cpp
```diff
@@ -40,6 +40,9 @@ int Pooling::load_param(const ParamDict& pd)
     global_pooling = pd.get(4, 0);
     pad_mode = pd.get(5, 0);
     avgpool_count_include_pad = pd.get(6, 0);
+    adaptive_pooling = pd.get(7, 0);
+    out_h = pd.get(8, 0);
+    out_w = pd.get(18, out_h);
```
nihui, line 45: The convention is small id for xxx_w, see kernel_w stride_w etc

nihui (changes requested):
for pooling_x86.cpp and pooling_arm.cpp

```
    if (adaptive_pooling)
    {
        return Pooling::forward(bottom_blob, top_blob, opt);
    }
```

nihui:
https://github.com/Tencent/ncnn/pull/2564 fix layer support checking

nihui (approved):

nihui:

PR #2565, opened 2021-01-07
pooling_x86_unused_opt

ncnnnnn:


src/layer/x86/pooling_x86.cpp
```diff
@@ -41,6 +41,7 @@ Pooling_x86::Pooling_x86()
 
 int Pooling_x86::create_pipeline(const Option& _opt)
 {
+    (void)_opt;
```
nihui, line 44: ```const Option& /*opt*/```

nihui (changes requested):

nihui:
Thanks !

PR #2566, opened 2021-01-07
add python build ci on mac and windows

caishanli:
add test mat, blob and paramdict
fix README.md build
print mat'data in python


nihui:
Thanks !

PR #2580, opened 2021-01-10
enhance padding test

nihui:


nihui:
Figure out why the driver crash when compiling padding_pack8to4.comp

PR #2621, opened 2021-01-25
fix iOS minimum version,when used Vukan 1.2

DaChengTechnology:
fix iOS minimum version,when used Vukan 1.2


nihui:
We will increase ios minimal deployment target version to 9.0 unconditionally

DaChengTechnology:
Update IOS mininum version to 9.0

nihui:

PR #2624, opened 2021-01-26
allow set llvm/mlir dirs from cli

daquexian:
`set` command with `CACHE` does not override the value passed from cli, so one can use `cmake -DLLVM_PROJECT_INSTALL_DIR=his_own_directory ..` to compile mlir2ncnn


nihui:
ok

PR #2633, opened 2021-01-27
upload wheel to pypi

caishanli:
use ncnn's version string as pyncnn's version
test release to pypi


python/setup.py.i
```diff
@@ -2,22 +2,43 @@ from setuptools import setup, find_packages
...
+  version           = "${PACKAGE_VERSION}",
+  author            = "nihui",
+  author_email      = "nihuini@tencent.com",
+  maintainer        = "caishanli",
+  maintainer_email  = "caishanli25@gmail.com",
+  description       = "ncnn is a high-performance neural network inference framework optimized for the mobile platform",
+  url               = "https://github.com/Tencent/ncnn",
+  classifiers       = [
+        "Programming Language :: Python :: 3",
+        "License :: OSI Approved :: MIT License",
```
nihui, line 35: bsd 3-clause ?

nihui (changes requested):

python/setup.py.i
caishanli: this is copied from xiaohanxiao's kears2ncnn......
```License :: OSI Approved :: BSD License```
is this correct?

caishanli: all license is in [classifiers](https://pypi.org/classifiers/)

nihui: BSD License is fine

PR #2645, opened 2021-02-01
add doc: use ncnn with own project

zchrissirhcz:
Hi Nihui

This PR creates a document, telling people how to integrate ncnn into their own project.


nihui:

PR #2651, opened 2021-02-02
workaround bugihfa cannot access fp16 element by vector index

zchrissirhcz:
Hi Nihui

This PR fix permute layer vulkan implementation by adding bugihfa related stuffs, as workaround for QCOM driver bug when acceccing vec4 element by dynamically calculated index.


nihui:
Thanks !

PR #2656, opened 2021-02-03
feat: add denormal options

leeys888:
Flush-To-Zero(FTZ) and Denormals-Are-Zero(DAZ) are modes that bypass IEEE754 methods of dealing with denormal floating-point numbers on x86_64 and some x86 CPUs. [[link]](https://en.wikipedia.org/wiki/Denormal_number#Disabling_denormal_floats_at_the_code_level)
refer : https://github.com/pytorch/pytorch/blob/22a34bcf4e5eaa348f0117c414c3dd760ec64b13/caffe2/core/init_denormals.cc

I haven't found a function in ARM chipset that does the following:
If someone finds it, please add it.

The following is the result of testing with a model with denormal number.
|                    | before | after |
|:------------------:|:------:|:-----:|
| custom mobilenetv2 |  28ms  |  8ms  |

### DIFF
--- a/src/cpu.cpp
+++ b/src/cpu.cpp
@@ -71,6 +71,10 @@
 #endif
 #endif
 
+#if defined(__SSE3__)
+#include <immintrin.h>
+#endif
+
 namespace ncnn {
 
 #if defined __ANDROID__ || defined __linux__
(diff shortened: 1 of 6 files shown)


src/cpu.cpp
```diff
@@ -847,4 +851,12 @@ void set_kmp_blocktime(int time_ms)
 #endif
 }
 
+void set_denormals_zero()
+{
+#if defined(__SSE3__)
+    _MM_SET_FLUSH_ZERO_MODE(_MM_FLUSH_ZERO_ON);
+    _MM_SET_DENORMALS_ZERO_MODE(_MM_DENORMALS_ZERO_ON);
+#endif
+}
```
nihui, line 860: please provide a parameter for controlling the Flush Zero and Denorm Zero mode, so that caller could restore the previous state.
It would be more convenient to have a getter function too.

benchmark/benchncnn.cpp
```diff
@@ -240,6 +240,8 @@ int main(int argc, char** argv)
     ncnn::set_omp_dynamic(0);
     ncnn::set_omp_num_threads(num_threads);
 
+    ncnn::set_denormals_zero();
```
nihui, line 243: It is fine to ask user to enable it explicitly.
I think we can do better for the api usage.

Add a switch to the Option class, and make it be true in constructor.
In Extractor::extract function, check the option flag and enable it when needed, and restore it after inference.
You may refer how the opt.openmp_blocktime flag is handled.
So, users would get the speed optimization automatically without changing any code.

nihui (changes requested):

src/cpu.cpp
leeys888: I agree to turn on/off FTZ and DAZ mode through parameters as your suggestion.
I'll add a getter and update it so that it can be set through parameters.

However, I am curious about the advantages of disabling FTZ or DAZ during the inference.

src/cpu.cpp
```diff
@@ -847,4 +851,49 @@ void set_kmp_blocktime(int time_ms)
 #endif
 }
 
+static int g_flush_denormals = 0;
```
nihui, line 854: The global variable will cause data-race in multi-threaded inference jobs. We always modify this state each time Extractor::extract() called, and multiple threads may call it concurrently.

Use ThreadLocalStorage in platform.h and set/get functions.

```
static ncnn::ThreadLocalStorage tls_flush_denormals;
```

some code samples
https://github.com/Tencent/ncnn/blob/9fd4d371ae03a70a73162ecde624c7feb8c0a82b/src/simpleomp.cpp#L294
https://github.com/Tencent/ncnn/blob/9fd4d371ae03a70a73162ecde624c7feb8c0a82b/src/simpleomp.cpp#L304

nihui (changes requested):

nihui (approved):

nihui:

PR #2681, opened 2021-02-17
support adaptive_pooling in vulkan implementation

zylo117:
support adaptive_pooling in vulkan implementation by adding a new shader based on the normal pooling shader but with adaptive kernel size and no padding.

Tested on my NVIDIA GTX1050 (mobile) and x86 cpu on win10 x64, the results are basically identical.

### DIFF
--- a/src/layer/vulkan/pooling_vulkan.cpp
+++ b/src/layer/vulkan/pooling_vulkan.cpp
@@ -388,8 +388,8 @@ int Pooling_vulkan::forward(const VkMat& bottom_blob, VkMat& top_blob, VkCompute
         constants[9].i = top_blob.cstep;
 
         const Pipeline* pipeline = elempack == 8 ? pipeline_pooling_global_pack8
-                                   : elempack == 4 ? pipeline_pooling_global_pack4
-                                   : pipeline_pooling_global;
+                                                 : elempack == 4 ? pipeline_pooling_global_pack4
+                                                                 : pipeline_pooling_global;
 
         cmd.record_pipeline(pipeline, bindings, constants, top_blob);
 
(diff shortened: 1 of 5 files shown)


src/layer/vulkan/pooling_vulkan.cpp
```diff
@@ -343,6 +396,40 @@ int Pooling_vulkan::forward(const VkMat& bottom_blob, VkMat& top_blob, VkCompute
         return 0;
     }
 
+    if (adaptive_pooling)
+    {
+        int outw = 1;
+        int outh = 1;
```
nihui, line 402: hardcode outw outh ?
There are parameterized out_w and out_h

nihui (changes requested):

src/layer/vulkan/pooling_vulkan.cpp
zylo117: fixed

src/layer/vulkan/shader/pooling_adaptive.comp
```diff
@@ -72,28 +72,34 @@ void main()
...
     afp res;
 
+    // calculate adaptive kernel size
+    int sx = int(floor(afp(gx * psc(w)) / afp(psc(outw))));
+    int ex = int(ceil(afp((gx + 1) * psc(w)) / afp(psc(outw))));
+    int kernel_w = ex - sx;
+    int sy = int(floor(afp(gy * psc(h)) / afp(psc(outh))));
+    int ey = int(ceil(afp((gy + 1) * psc(h)) / afp(psc(outh))));
+    int kernel_h = ey - sy;
+
```
nihui, line 82: What about resolving the kernel size on cpu side, and passing them via push constants ?

zylo117: But according to the implementation of adaptive pooling in c++ and pytorch, the kernel size is dynamic and calculated on every location. 

For example, input size and output size is 5 and 3, the kernel size for each data will be 2, 3, 2. 

Can this be done on cpu side(pooling_vulkan.cpp) since kernel size is no long constant?

ncnn c++:
https://github.com/Tencent/ncnn/blob/ab56083ca543d4315bef9913b23b843e1fdd8c76/src/layer/pooling.cpp#L114-L128

pytorch:
https://github.com/pytorch/pytorch/blob/51861cc9b19d9c483598e39932661822a826d3a2/aten/src/ATen/native/AdaptiveAveragePooling.cpp#L34-L52

nihui: okay !

nihui (approved):

nihui:

PR #2689, opened 2021-02-19
feat(ncnnoptimize): replace denormals to zero on layers with weights

leeys888:


nihui:
How about modifying these two serialization functions ?
All the layers with weight data will call them for writing weight data to file.
So we don't need to handle the difference of layer types.

In addition, I think replacing denormals to zero will bring little negative effects. It is not necessary to make another command line option for disabling it.
```
NetOptimize::fwrite_weight_tag_data
NetOptimize::fwrite_weight_data
```

leeys888:
> How about modifying these two serialization functions ?
> All the layers with weight data will call them for writing weight data to file.
> So we don't need to handle the difference of layer types.
> 
> In addition, I think replacing denormals to zero will bring little negative effects. It is not necessary to make another command line option for disabling it.
> 
> ```
> NetOptimize::fwrite_weight_tag_data
> NetOptimize::fwrite_weight_data
> ```

Oh! Thank you for your suggestion.
I think thats a good idea.

`NetOptimize::fwrite_weight_tag_data` `NetOptimize::fwrite_weight_data` 
I'll update to replace denormals through the above two functions. :D

nihui:
Hi, thanks for your work.
Could you please add another function for replacing fp16 denormals ?

leeys888:
> Hi, thanks for your work.
> Could you please add another function for replacing fp16 denormals ?

@nihui 
Hi, I think the fp16 denormal function is applied to the function below. Please check once.
https://github.com/Tencent/ncnn/blob/4083eafbc0822028f003164d8b03301ec1747a53/tools/ncnnoptimize.cpp#L2979
https://github.com/Tencent/ncnn/blob/4083eafbc0822028f003164d8b03301ec1747a53/src/mat.cpp#L197

nihui:
> > Hi, thanks for your work.
> > Could you please add another function for replacing fp16 denormals ?
> 
> @nihui
> Hi, I think the fp16 denormal function is applied to the function below. Please check once.
> 
> https://github.com/Tencent/ncnn/blob/4083eafbc0822028f003164d8b03301ec1747a53/tools/ncnnoptimize.cpp#L2979
> 
> https://github.com/Tencent/ncnn/blob/4083eafbc0822028f003164d8b03301ec1747a53/src/mat.cpp#L197

yeah, all the fp32 denormals will be zero when converting to fp16
However, there are some normal fp32 that cannot be represented as normal fp16. They will be denormal fp16.

https://github.com/Tencent/ncnn/blob/4083eafbc0822028f003164d8b03301ec1747a53/src/mat.cpp#L241

leeys888:
> > > Hi, thanks for your work.
> > > Could you please add another function for replacing fp16 denormals ?
> > 
> > 
> > @nihui
> > Hi, I think the fp16 denormal function is applied to the function below. Please check once.
> > https://github.com/Tencent/ncnn/blob/4083eafbc0822028f003164d8b03301ec1747a53/tools/ncnnoptimize.cpp#L2979
> > 
> > https://github.com/Tencent/ncnn/blob/4083eafbc0822028f003164d8b03301ec1747a53/src/mat.cpp#L197
> 
> yeah, all the fp32 denormals will be zero when converting to fp16
> However, there are some normal fp32 that cannot be represented as normal fp16. They will be denormal fp16.
> 
> https://github.com/Tencent/ncnn/blob/4083eafbc0822028f003164d8b03301ec1747a53/src/mat.cpp#L241

Thank you. I understood it.
I'll implement it in `unsigned short float32_to_float16(float value);`

tencent-adm:
[CLA assistant check](https://cla.opensource.tencent.com/Tencent/ncnn?pullRequest=2689) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.opensource.tencent.com/Tencent/ncnn?pullRequest=2689) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.opensource.tencent.com/check/Tencent/ncnn?pullRequest=2689) it.</sub>

nihui (approved):

nihui:
Thanks !

nihui:
merged as https://github.com/Tencent/ncnn/pull/2690

PR #2691, opened 2021-02-20
fix a missing import in python setup.py

zylo117:
fix a missing import in python setup.py or python setup will fail


nihui:
Thanks !

PR #2697, opened 2021-02-21
fix avg pooling accumulation overflow in vulkan using fp16 arithmetic

zylo117:
The current implementation use fp16 variable to perform accumulation in avg pooling, 

for example,
```
res = afp(0.f); // res will be a fp16 variable in fp16 mode
for (...)
    res += buffer_ld1(bottom_blob_data, v_offset + x); // res keeps getting larger
```
So there is a protenial risk of data overflow considering `res` keeps getting larger over time.

Assuming there is a feature map with a larger kernel size 256x256, and the feature map has value over `1.0` on each data, the sum will be over `65536`, causing fp16 overflow, then you get the wrong result of pooling.

I also confirmed that when disabling all the fp16 options in vulkan, the result is correct and when running in vulkan on older devices which don't support fp16 arithmetic like gtx1050, the result is also right.

Now I perform accumulation on a temporary fp32 variable and cast to fp16 in the end. Though it may be a little bit slower but it can make sure the result is accurate. It's tested on rtx2080ti and intel u630, the result is correct in both fp16 and fp32 modes.

### DIFF
--- a/src/layer/vulkan/shader/pooling.comp
+++ b/src/layer/vulkan/shader/pooling.comp
@@ -119,7 +119,7 @@ void main()
     }
     if (pooling_type == 1 && avgpool_count_include_pad == 0)
     {
-        res = afp(0.f);
+        float res_fp32 = 0.f;  // force accumulation in fp32
         int area = 0;
 
         int sx = gx * stride_w;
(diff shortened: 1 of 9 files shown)


nihui:
What about using fp32 accumulator for global pooling and adaptive pooling only ?
For normal pooling, there is little chance of very large kernel size.

zylo117:
yes, I'm working on it

PR #2708, opened 2021-02-25
remove unnecessary mat clone when NCNN_BENCHMARK enabled

yx9527:


nihui:

PR #2709, opened 2021-02-25
fix that inplace forward never happened

gdh1995:
During `src/net.cpp#NetPrivate::do_forward_layer`, The old code always creates a clone of `blob_mats[bottom_blob_index]`, so its `refcount` is always larger than or equal with 2.

Then this PR fixes it by delaying shadow-cloning those bottom mat instance.


nihui:

nihui (approved):

PR #2722, opened 2021-03-02
fix array index out of bounds in examples/yolact.cpp

cmdbug:
ncnn yes, nihui yes.


nihui (approved):

nihui:
Thanks for your contribution!

PR #2732, opened 2021-03-06
fix net and extractor destroy order when use vulkan

caishanli:


src/net.cpp
```diff
@@ -2195,13 +2195,24 @@ class ExtractorPrivate
...
+        ,
+          vkdev(0)
+#endif // NCNN_VULKAN
     {
+#if NCNN_VULKAN
+        if (net)
+        {
+            vkdev = net->vulkan_device();
+        }
+#endif // NCNN_VULKAN
```
nihui, line 2208: move member initialization to ```Extractor::Extractor(const Net* _net, size_t blob_count)```

nihui (changes requested):

nihui (approved):

nihui:
谢谢你

PR #2762, opened 2021-03-17
add net Cut Function

chentyjpm:
use layername instead of index

add net Cut Function
cut param and bin form start to end

usage: %s [inparam] [inbin] [outparam] [outbin] [flag] [cutstart] [cutend]

compatible
if no [cutstart] [cutend] will work as old cmd optimise all network


nihui:
merged in https://github.com/Tencent/ncnn/pull/2763

Issue #2826, opened 2021-04-08
Hardswish is not able to produce correct results


MarsTechHAN:
For the default parameters, the result is not able to match with the baseline.
![image](https://user-images.githubusercontent.com/10462805/114063197-f075cd80-98ca-11eb-9c92-eb1e4ce57ced.png)

Code to reproduce the result:
```python
import numpy as np
import matplotlib.pyplot as plt

alpha = 0.2
beta = 0.5

lower = -alpha / beta
upper = (1 / alpha) + lower

print('Upper: %.03f, Lower: %.03f' % (upper, lower))
x = np.linspace(-10, 10, 100000, endpoint=True)

y_n = []
y_b = []

for i in range(len(x)):
    if (x[i] < lower):
        y_n.append(0.0)
    else:
        if (x[i] > upper):
            y_n.append(x[i])
        else:
            y_n.append(x[i] * (x[i] * alpha + beta))
    
    y_b.append(max(0, min(x[i] + 3, 6)) / 6 * x[i])

plt.plot(x, y_n, label = "ncnn alg")
plt.plot(x, y_b, label = "baseline")
plt.legend()
plt.savefig("x.png")
```

In order for the formula to produce the correct result
```python
alpha = 1/6
beta = 1/2

lower = -3
upper = 3
```
which is non-achievable for the layer:
https://github.com/Tencent/ncnn/blob/c4700c52ca0a8e107068691ac32f60886333f5d2/src/layer/hardswish.cpp#L27-L30

MarsTechHAN:
It seems like there is a mistake in my test program, after fixing it, only the alpha and beta need some adjustments, to 
```
alpha = 1/6
beta = 1/2
```
![image](https://user-images.githubusercontent.com/10462805/114065399-359aff00-98cd-11eb-9bfb-945ae0fdffa8.png)

songqun:
Hi @Howave , could you help check the value of alpha/beta in hardswish?
From the formula: `x*relu6(x+3)/6`, the quadratic function part should be `(x^2)/6 + x/2` and alpha/beta should be 1/6 and 1/2. 
Is there any exact reason to set them as 0.2 and 0.5?

MarsTechHAN:
HardSigmoid also does not have the correct paremeters:
https://github.com/Tencent/ncnn/blob/c4700c52ca0a8e107068691ac32f60886333f5d2/src/layer/hardsigmoid.cpp#L25-L59

PR #2859, opened 2021-04-21
fix default alpha beta in hardswish and add comment for them in hardsigmod

songqun:
Close https://github.com/Tencent/ncnn/issues/2826

### DIFF
--- a/src/layer/hardswish.cpp
+++ b/src/layer/hardswish.cpp
@@ -24,6 +24,8 @@ HardSwish::HardSwish()
 
 int HardSwish::load_param(const ParamDict& pd)
 {
+    // Note that tensorflow/pytorch use alpha,beta = 1/6, 0.5, not the default value here.
+    // You can setup them manually in .param file.
     alpha = pd.get(0, 0.2f);
     beta = pd.get(1, 0.5f);
     lower = -beta / alpha;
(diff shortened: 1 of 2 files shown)


src/layer/hardswish.cpp
```diff
@@ -24,7 +24,7 @@ HardSwish::HardSwish()
 
 int HardSwish::load_param(const ParamDict& pd)
 {
-    alpha = pd.get(0, 0.2f);
+    alpha = pd.get(0, 1.f/6);
```
songqun, line 27: not sure whether we need 0.1666667f instead.

songqun:
Hi @nihui, could you help review the MR...

src/layer/hardswish.cpp
```diff
@@ -24,7 +24,7 @@ HardSwish::HardSwish()
 
 int HardSwish::load_param(const ParamDict& pd)
 {
-    alpha = pd.get(0, 0.2f);
+    alpha = pd.get(0, 1.f / 6);
```
nihui, line 27: This breaks backward compatibility.
Suppose there is one old model file using the default alpha value. You will get different output with or w/o this change.

nihui (changes requested):

src/layer/hardswish.cpp
songqun: Have reverted it and add some comments instead.

nihui (approved):

nihui:

PR #2972, opened 2021-06-02
test release python on s390x

caishanli:


nihui:
python wheel now supports s390x  :D

PR #3011, opened 2021-06-18
Update build for visual studio Chinese doc

zchrissirhcz:
1. Add `""` for `%cd%/install`
2. Providing `-G Visual Studio 16 2019` as alternative to `nmake` generator
3. Providing simple build steps that don't require protobuf 
4. Providing vulkan based build steps
5. Mention build steps for mlir2ncnn's link
6. Mention `find package ncnn` and provide link


nihui:
Thanks !

PR #3087, opened 2021-07-16
darknet2ncnn: fix convolution padding when pad=0

zhiliu6:


nihui:

PR #3090, opened 2021-07-18
fix AVX2 compiled as AVX problem

zhiliu6:


nihui:

PR #3097, opened 2021-07-20
RVV optimized BinaryOp

thelastlin:
RISC-V "V" extension optimized BinaryOp, with fp16 support.


nihui:

PR #3098, opened 2021-07-21
Add resize_nearest definition

AP-Kai:
Add resize_nearest definition


zchrissirhcz:
The added APIs, looks good to me.
But the implementations are still empty.

I would recommend change this PR to draft PR thus can continously push commit and be seen here.

QiuKaiyyds:
> The added APIs, looks good to me.
> But the implementations are still empty.
> 
> I would recommend change this PR to draft PR thus can continously push commit and be seen here.

Thanks for your advises! I will improve the PR as soon as possible.

PR #3107, opened 2021-07-24
Update openmp installation methods and M1 benchmarks

zhiqwang:
Hi @nihui 

I updated the `openmp` installation instructions and M1 benchmarks in this PR.


nihui:

PR #3108, opened 2021-07-25
RVV optimized HardSwish and HardSigmoid

thelastlin:
RVV optimized HardSwish and HardSigmoid, with fp16 support.


nihui:

PR #3113, opened 2021-07-28
RVV optimized PReLU

thelastlin:
RVV optimized PReLU, with fp16 support


nihui:

PR #3114, opened 2021-07-28
Add unittest and RVV optimized for  SELU

thelastlin:


nihui:

PR #3164, opened 2021-08-11
Update RK3328 CPU benchmark

Liuyufanlyf:


nihui:

dichild:
Thanks a lot!!!   I also try  building  ncnn on R2S, but failed with a libc problem. What firmware did you use? It would be great if you share some infos about the build process or the system.

Liuyufanlyf:
> Thanks a lot!!! I also try building ncnn on R2S, but failed with a libc problem. What firmware did you use? It would be great if you share some infos about the build process or the system.

Well, I just cloned the latest code in branch master and followed "how to build" in the wiki to build the project. It took more than 30 minutes but finally succeeded. I'm using Armbian focal, maybe you are using a different OS.

PR #3172, opened 2021-08-13
Update Khadas Vim3 benchmark(add yolox model)

FeiGeChuanShu:


nihui:

PR #3214, opened 2021-09-09
fix hardswish test beta param

zhiliu6:


nihui:

PR #3254, opened 2021-09-27
Fix broken codepaths with AVX only

Timen:
Fix a couple of broken codepaths involving opt.use_weight_fp16_storage, when only AVX is enabled and not AVX2. This was not caught in tests as the testutil actually applied some overrides on the Option object to prevent enabling certain functions when the layer doesn't support it. 

I think this shouldn't happen as users may also enable Option features even though some layers might not support it, as such I removed that code from testutil.h.

### DIFF
--- a/.github/workflows/linux-x64-cpu-gcc.yml
+++ b/.github/workflows/linux-x64-cpu-gcc.yml
@@ -28,7 +28,7 @@ on:
     - 'examples/**'
 jobs:
   linux-gcc:
-    runs-on: ubuntu-16.04
+    runs-on: ubuntu-18.04
     steps:
     - name: cancel-previous-runs
       uses: styfle/cancel-workflow-action@0.9.1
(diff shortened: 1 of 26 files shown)


Timen:
The vulkan 1.114.0 SDK URL used in the windows build is broken and I changed it to 1.121.2

Timen:
Seems the only test that is failing now that I disabled the option overrides seem to be `test_padding` for a couple of targets(and `test_lrn` but that seems related). I will try and fix these soon but probably won't have time. @nihui Is this something you can help with? Overal I think disabling the option overrides is a good idea as many users will probably not realize some layers don't support an option but nothing is stopping them from enabling that specific option.

nihui:
issue inside padding implementation confirmed

nihui:
gpu test failures indicates the missing image storage support

Timen:
Thanks @nihui!

PR #3294, opened 2021-10-20
move custom_layer_to_index to public

Timen:
Small change to expose the custom_layer_to_index to public which is use-full when working with registered custom layers.


nihui:

PR #3301, opened 2021-10-25
get_elf_hwcap_from_getauxval

nihui:


nihui:
> @nihui Do you have any idea what is the cause of the flaky windows CI tests (which also failed in this completely unrelated PR) which produces the following error:
> 
> ```
> 47/70 Test #45: test_padding .....................***Failed    0.31 sec
> CMake Error at D:/a/ncnn/ncnn/cmake/run_test.cmake:4 (message):
>   Test failed with return value 'Access violation'
> ```
> 
> I saw it in a PR i did earlier as-well, and it's rather concerning.

unable to reproduce at the moment, and I have no idea how this happens  :(

PR #3333, opened 2021-11-08
use int ceil/floor div

Yutyrannus:


src/layer/pooling.cpp
```diff
@@ -119,12 +118,16 @@ int Pooling::forward(const Mat& bottom_blob, Mat& top_blob, const Option& opt) c
 
                 for (int i = 0; i < out_h; i++)
                 {
-                    const int ih0 = floor(((float)h / out_h) * i);
-                    const int ih1 = ceil(((float)h / out_h) * (i + 1));
+                    // floor div
+                    const int ih0 = h / out_h * i;
+                    // ceil div
+                    const int ih1 = (h + out_h - 1) / out_h * (i + 1);
```
nihui, line 124:
```
                    // floor div
                    const int ih0 = h * i / out_h;
                    // ceil div
                    const int ih1 = (h * (i + 1) + out_h - 1) / out_h;
```

nihui (changes requested):
and please fix the remaining

src/layer/pooling.cpp
Yutyrannus: fix

nihui (approved):

nihui:

PR #3345, opened 2021-11-14
Update operators.md

soragotosann:
change weight type


nihui:

PR #3395, opened 2021-12-02
Fix draw on const cv::Mat bug in p2pnet example

FeiGeChuanShu:


nihui (approved):

nihui:

PR #3408, opened 2021-12-08
dynamic convolution weight

nihui:


src/layer/convolution.cpp
```diff
@@ -176,103 +262,96 @@ int Convolution::forward(const Mat& bottom_blob, Mat& top_blob, const Option& op
...
-                if (bias_term)
-                    sum = bias_data[p];
+int Convolution::forward(const std::vector<Mat>& bottom_blobs, std::vector<Mat>& top_blobs, const Option& opt) const
+{
+    const Mat& bottom_blob = bottom_blobs[0];
+    const Mat& _weight_data = bottom_blobs[1];
+    Mat& top_blob = top_blobs[0];
 
-                const float* kptr = (const float*)weight_data + maxk * channels * p;
+    const int _kernel_w = _weight_data.w;
```
zchrissirhcz, line 299: A small suggestion: Local variable names starting with `_` may be shadowed by same name global variables, since they are reserved.
Reference: https://stackoverflow.com/a/228797/2999096

PR #3413, opened 2021-12-11
[pnnx]: auto copy dll files and remove Torch_INSTALL_DIR option in CI

Yutyrannus:


nihui (approved):

nihui:

PR #3421, opened 2021-12-13
add sw_64 platform support

wzyforgit:
Add "sw_64" platform support
All unit test have passed
Models that passed test: PaddleOCR v2.0, MobileNet v3, SqueezeNet v1.1
My CPU is "sw421"


nihui (approved):

nihui:

PR #3443, opened 2021-12-23
binary4d, unary4d

nihui:


nihui:
TODO upgrade riscv-gcc to https://github.com/riscv-collab/riscv-gcc/tree/riscv-gcc-10.1-rvv-dev

PR #3453, opened 2021-12-27
update rvv toolchain

nihui:


nihui:
qemu part merged in https://github.com/Tencent/ncnn/pull/3756
toolchain is not working currently :(

PR #3470, opened 2021-12-31
add the param file of yolo-fastest in benchmark

dog-qiuqiu:
add the param file of yolo-fastest in benchmark


nihui:

PR #3527, opened 2022-01-24
x86/avx_mathfun.h: Remove fallback warnings

kagurazakakotori:
This patch removes warning messages indicating falling back to SSE2
when AVX2 support is disabled as suggested. Also reorders non-AVX2
macros for readability and faster preprocessing.

Suggested-by: nihui <shuizhuyuanluo@126.com>
Signed-off-by: Kagurazaka Kotori <kagurazakakotori@gmail.com>

### DIFF
--- a/src/layer/x86/avx_mathfun.h
+++ b/src/layer/x86/avx_mathfun.h
@@ -335,7 +335,7 @@ _PS256_CONST(cephes_FOPI, 1.27323954473516f); // 4 / M_PI
 
 */
 static NCNN_FORCEINLINE __m256 sin256_ps(__m256 x)
-{ // any x
+{   // any x
     __m256 xmm1, xmm2 = _mm256_setzero_ps(), xmm3, sign_bit, y;
     __m256i imm0, imm2;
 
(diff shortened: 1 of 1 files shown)


kagurazakakotori:
Tested only with non-AVX2 build as my build server does not support AVX2 (Intel Ivy Bridge-EP Xeon E5). All unit testes passes with some modification in `test_squeezenet`.

Seems the relative path of the files to be loaded is wrong, it should be `../examples`.

https://github.com/Tencent/ncnn/blob/2d46994d2e992753b8982c1a929610589c2b2d8d/tests/test_squeezenet.cpp#L170

kagurazakakotori:
> Tested only with non-AVX2 build as my build server does not support AVX2 (Intel Ivy Bridge-EP Xeon E5). All unit testes passes with some modification in `test_squeezenet`.
> 
> Seems the relative path of the files to be loaded is wrong, it should be `../examples`.
> 
> https://github.com/Tencent/ncnn/blob/2d46994d2e992753b8982c1a929610589c2b2d8d/tests/test_squeezenet.cpp#L170

My build setup issue, please ignore this.

nihui (approved):

nihui:

PR #3536, opened 2022-01-28
x86: Use _mm_cvtsi128_si{32,64} in float2int8

kagurazakakotori:
This patch uses _mm_cvtsi128_si{32,64} intrinsics when returning value
in float2int8() to reduce unnecessary memory accesses.

Resolves TODO "use _mm_cvtsi128_si64 on 64bit target".

Signed-off-by: Kagurazaka Kotori <kagurazakakotori@gmail.com>


kagurazakakotori:
All unit tests passes on non-AVX2 environment (Intel Ivy Bridge Xeon E5)

nihui:

PR #3540, opened 2022-02-01
Optimize scale x86

Yoh-Z:
Complete the x86 scale code and passed the local test.

### DIFF
--- a/src/layer/x86/scale_x86.cpp
+++ b/src/layer/x86/scale_x86.cpp
@@ -35,178 +35,98 @@ int Scale_x86::forward_inplace(std::vector<Mat>& bottom_top_blobs, const Option&
     Mat& bottom_top_blob = bottom_top_blobs[0];
     const Mat& scale_blob = bottom_top_blobs[1];
 
+    int w = bottom_top_blob.w;
+    int h = bottom_top_blob.h;
+    int channels = bottom_top_blob.c;
     int dims = bottom_top_blob.dims;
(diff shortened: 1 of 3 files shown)


src/layer/x86/scale_x86.cpp
```diff
@@ -35,53 +35,92 @@ int Scale_x86::forward_inplace(std::vector<Mat>& bottom_top_blobs, const Option&
...
+            else
+            {
+                *ptr = *ptr * *scale;
             }
+
+            ptr++, scale++;
         }
 
+        return 0;
+    }
```
nihui, line 117: for dim==1 case, move bias condition out of the loop

data race happens in ptr and scale pointer with multi-threading

nihui (changes requested):

src/layer/x86/scale_x86.cpp
```diff
@@ -298,97 +282,194 @@ int Scale_x86::forward_inplace(std::vector<Mat>& bottom_top_blobs, const Option&
...
             }
         }
+
+        return 0;
     }
 
-    return 0;
+#endif // __SSE2__
+
+    return Scale::forward_inplace(bottom_top_blobs, opt);
```
nihui, line 472: do not wrap elempack=1 cases inside sse2 macro

nihui (changes requested):

nihui (approved):

nihui:

PR #3568, opened 2022-02-16
drop x86 avx2 fp16

nihui:


Timen:
Sad to see it go :( But it doesn't make sense to maintain it, as it's very niche to use.

nihui:
> Sad to see it go :( But it doesn't make sense to maintain it, as it's very niche to use.

we now have much better sse2/avx optimization now  :P

PR #3592, opened 2022-03-04
Add arm and arm64 targets support for MSVC.

MouriNaruto:
I was surprised to find that ncnn cannot be compiled properly in arm and arm64 targets with MSVC toolchain yesterday. So, I think it's time to add them for ncnn.

The related logic is so easy for arm and arm64 targets with MSVC toolchain because it seems that MSVC arm and arm64 compiler don't support NCNN_ARM82 and NCNN_ARM82DOT.

Currently, I haven't enabled the Vulkan support for ncnn because the Vulkan SDK don't provide the arm and arm64 Windows runtime binaries. I haven't enabled the protobuf support because of the similar reason. (But I think we may build protobuf arm and arm64 Windows binaries with vcpkg properly.)

Have a good time!

Kenji Mouri


nihui (approved):

nihui:

PR #3595, opened 2022-03-04
Add SSE2 implementation of floor and ceil in x86 targets.

MouriNaruto:
Just do some implementations but I don't understand why because I dump the implementations from compiled binaries.

P.S. floor implementation has bug. Give me some time to fix.

Have a good time!

Kenji Mouri


nihui:

PR #3623, opened 2022-03-15
style(src/layer): remove unused para and var

tpoisonooo:
To prevent unpredictable crash/error, such as [float precision problem](https://github.com/tpoisonooo/cpp-syntactic-sugar/tree/master/float-precision), we must treat warning as error...

This issue reomove unused parameter and var during compiling.


tpoisonooo:
看起来那个 intrinsic kernel 没啥用。

nihui:
lgtm

nihui:

PR #3651, opened 2022-03-24
Update convolution_vulkan.cpp

tpoisonooo:
fix build warning `-Wparentheses`.


nihui:

PR #3656, opened 2022-03-26
Update yolov5.py

hariag:
fix rect position bug.


nihui (approved):

nihui:

PR #3726, opened 2022-04-23
RVV Optimized erfc and GELU

thelastlin:


nihui:

PR #3733, opened 2022-04-24
Add the benchmark of Jetson AGX Orin

BUG1989:


nihui:

PR #3753, opened 2022-04-30
RVV: use vfredu(o)sum.vs as vfredsum.vs alts

thelastlin:
The latest RISC-V Vector intrinsic spec uses `vfredusum` to replace `vfredsum`.

This PR is a simple workaround that detects which intrinsic function is provided by the compiler, and, then appends a macro to replace `vfredsum` with `vfredusum` or `vfredosum` at configure stage, without changing any other RVV layer src.

It's  recommended that replacing the `vfredsum` with `vfredusum` in RVV Layer src with a header that can replace `vfredusum` with `vfredsum` at build stage with the help of CMake.


CMakeLists.txt
```diff
@@ -194,6 +194,21 @@ elseif(CMAKE_SYSTEM_PROCESSOR MATCHES "^(riscv)")
     check_cxx_compiler_flag("-march=rv64gcv_zfh" NCNN_COMPILER_SUPPORT_RVV_FP16)
 
     if(NCNN_COMPILER_SUPPORT_RVV)
+        include(${CMAKE_CURRENT_SOURCE_DIR}/cmake/ncnn_check_rvv_vfredsum.cmake)
+        if(NOT NCNN_COMPILER_USE_VFREDSUM AND NCNN_COMPILER_SUPPORT_RVV_INTRINSIC)
+            if(NCNN_COMPILER_USE_VFREDUSUM)
+                SET(NCNN_REPLACED_VFREDSUM vfredusum)
+            else()
+                SET(NCNN_REPLACED_VFREDSUM vfredosum)
```
thelastlin, line 202: It seems that some assemblers are unable to convert `vfredusum` into corresponding binary.

thelastlin:
Replaced by #3790

PR #3781, opened 2022-05-12
pass build

cugxchen:


nihui:
duplicate of https://github.com/Tencent/ncnn/pull/3779

PR #3790, opened 2022-05-14
RVV: follow intrinsic doc, replace vfredsum_* with vfredusum_*

thelastlin:
The latest RISC-V Vector intrinsic spec uses vfredusum to replace vfredsum.
This PR replace vfredsum_* with corresponding vfredusum_*.

This PR contains a simple workaround that detects which intrinsic function is provided by the compiler, and, then appends a macro to replace vfredusum with vfrdusum at configure stage and warns users to upgrade toolchains.


nihui:

PR #3813, opened 2022-05-20
Add the benchmark of AX620A

BUG1989:
Add the benchmark of AX620A with CPUs.


nihui:

PR #3838, opened 2022-05-24
Fix x86 unaryop

Yoh-Z:


src/layer/x86/unaryop_x86.cpp
```diff
@@ -66,56 +65,56 @@ static int unary_op_inplace(Mat& a, const Option& opt)
...
+            *ptr = op.func(*ptr);
             ptr++;
         }
     }
 
     return 0;
 }
 
-namespace UnaryOp_x86_functor {
+namespace UnaryOp_x86_avx_functor {
```
nihui, line 99: do not change namespace

src/layer/x86/unaryop_x86.cpp
```diff
@@ -464,24 +463,47 @@ struct unary_op_cos
...
+    __m512 func_pack16(const __m512& x) const
     {
-        return tan512_ps(x);
+        //TODO avx512 optimize
+        float tmp[16];
+        _mm512_storeu_ps(tmp, x);
+        for (int i = 0; i < 16; i++)
+            tmp[i] = tan(tmp[i]);
+        return _mm512_loadu_ps(tmp);
     }
```
nihui, line 507: tan optimization lost ?

nihui (changes requested):

nihui (approved):

nihui:

PR #3839, opened 2022-05-25
fix yolox input shape w!=h

FeiGeChuanShu:


nihui:

PR #3868, opened 2022-06-01
fix loongarch64 ci

nihui:


nihui:
upstream issue https://github.com/loongson/build-tools/issues/18

PR #3893, opened 2022-06-07
Added a simple unittest for Power layer.

proydakov:
Hi, NCNN Team.

As I see from code coverage report we don't have a unittest for Power layer:

https://app.codecov.io/gh/Tencent/ncnn/blob/master/src/layer/power.cpp

I have added a simple test_power.cpp and updated CMakeLists.txt.

Could you review & accept my change, pls?

Best regards, Evgeny Proydakov.


nihui:
Hi, thanks for your contribution

Power layer has three optional parameters
https://github.com/Tencent/ncnn/blob/master/docs/developer-guide/operators.md#power
Please set these parameters to test the case of non-default parameters

proydakov:
Let me check.

proydakov:
I have updated params in unittes.

static int test_power(const ncnn::Mat& a)
{
    ncnn::ParamDict pd;
    pd.set(0, 1.1f);
    pd.set(1, 1.5f);
    pd.set(2, 2.0f);

nihui:

PR #3983, opened 2022-07-02
fix(examples): exit program if load_param, load_model failed for clear message

zchrissirhcz:


zchrissirhcz:
For https://github.com/Tencent/ncnn/issues/3982

nihui (changes requested):
load function returns 0 if success
thus, the condition shall check any non-zero value for any potential failure

zchrissirhcz:
Checking zero/nonzero now.
My rebase failed to push to origin, thus made an new commit.

nihui:

PR #3988, opened 2022-07-04
pnnx convert torch.tensor_split

nihui:


tools/pnnx/tests/ncnn/test_torch_tensor_split.py
```diff
@@ -0,0 +1,63 @@
...
+import torch.nn.functional as F
+
+class Model(nn.Module):
+    def __init__(self):
+        super(Model, self).__init__()
+
+    def forward(self, x, y, z):
+        x0, x1, x2 = torch.tensor_split(x, (12, 13))
+        y0, y1, y2 = torch.tensor_split(y, 3, dim=1)
+        z0, z1 = torch.tensor_split(z, (3,), dim=0)
```
csukuangfj, line 26: please also add tests for negative indexes, i.e., using
```python3
z0, z1 = torch.tensor_split(z, (-3,), dim=0)
```
and you will see the test will fail.

PR #3989, opened 2022-07-04
[wip][rfc] C api refactor

djdisodo:
issue i faced:

i was making rust bindings
tried to find a way to extend types like allocator

correct me if i'm wrong, while current design allows user to change function it doesn't able the function to access user's data in reliable way

(for allocator, and more classes)

i suggest to create a class that extends allocator by wrapping data pointer and member function implementation from c

more things:

typedef types like `ncnn_allocator_t` "as is" when it's cpp

example:

```cpp
/* allocator api */
#ifdef __cplusplus
typedef Allocator* ncnn_allocator_t;
#else
typedef struct __ncnn_allocator_t* ncnn_allocator_t;
#endif
```

this will prevent redundant type castings

[https://github.com/Tencent/ncnn/blob/b4bae2c9e41ac8a6d686ef20c3bbd5a217717115/src/c_api.cpp#L754](https://github.com/Tencent/ncnn/blob/b4bae2c9e41ac8a6d686ef20c3bbd5a217717115/src/c_api.cpp#L754)

code below converts Mat*[] into Mat[]
we can prevent this conversion by exposing struct Mat to c

i'm not sure but ModelBinFromMatArray seems to hold reference of matarray(which seems to be dropped at return)
is it safe to ?
```cpp
ncnn_modelbin_t ncnn_modelbin_create_from_mat_array(const ncnn_mat_t* weights, int n)
{
    std::vector<Mat> matarray(n);
    for (int i = 0; i < n; i++)
    {
        matarray[i] = *(const Mat*)weights[i];
    }
    ncnn_modelbin_t mb = (ncnn_modelbin_t)malloc(sizeof(struct __ncnn_modelbin_t));
    mb->pthis = (void*)(new ModelBinFromMatArray_c_api(mb, &matarray[0]));
    mb->load_1d = __ncnn_ModelBinFromMatArray_load_1d;
    mb->load_2d = __ncnn_ModelBinFromMatArray_load_2d;
    mb->load_3d = __ncnn_ModelBinFromMatArray_load_3d;
    return mb;
}
```

so i suggest exposing Mat and Option(for easier use) to c
by changing them to structs and changing mat.h and option.h to be c compatible(by defining member functions only when it's cpp)

Option is already almost "struct"
(there's currently no difference between struct and class in cpp tho)

also, to not expose c api to cpp much even if i define c api structs in mat.h
i suggest putting c structs and other c api functions into ncnn::c_api namespace

...i'd like to move even more functions from c_api.h to current cpp headers


nihui:
Hi, thanks for the comments about the usage of c_api

ncnn uses the c++ interface, and c_api is a thin wrapper of the cpp interface, which is used to deal with some special scenarios where the c++ interface cannot be used.
Incorporating C-style interfaces into C++ code leads to confusion, don't do it.

We can add `void* user_data` parameters to some c interfaces, such as `void* (*fast_malloc)(ncnn_allocator_t allocator, size_t size, void* user_data);`, user_data is passed in when the structure is created, which allows the caller to interact with its own data , which is a common practice in many c libraries.

djdisodo:
@nihui 
https://github.com/Tencent/ncnn/blob/dec6622ec4ae4ad866f6b04c392369c825afe120/src/c_api.cpp#L54
allow extending cpp class like this



https://github.com/Tencent/ncnn/blob/dec6622ec4ae4ad866f6b04c392369c825afe120/src/c_types.h#L1
added c_types(type alias) like this



https://github.com/Tencent/ncnn/blob/dec6622ec4ae4ad866f6b04c392369c825afe120/src/mat.h#L343
using c style type here



putting Allocator* instead of ncnn_allocator_t will involve a bit more complex macros
it's same type anyway
should i?

djdisodo:
https://github.com/Tencent/ncnn/blob/7886e90c65ec995159427e9e98bf9520683f661e/src/c_api.h#L186
seems like this allows to change Layer's behavior
but only when pthis is Layer_c_api
which leads to confusion

since Layer has not only virtual functions but also fields and it's public
i don't think there's clear way to allow c to extend not only Layer but other classes extending Layer without writing bindings for each of them
(i'm not sure but we can probably make some class like ILayer which only has virtual functions and make Layer to extend/inherit it)

PR #3992, opened 2022-07-05
Add Loongson 2F toolchain support (refer to AOSC)

luyanaa:
Loongson 2F's mainline toolchain needs "-mabi=64 -march=mips3" to work. And add -mloongson-mmi support in toolchain.


nihui:
hi, is there any ncnn benchmark result on loongson2f ?

PR #4034, opened 2022-07-14
armv8.4 i8mm optimization for convolution gemm int8

nihui:


nihui:
TODO: code clean and add some comment

PR #4060, opened 2022-07-21
[LayerNorm Optimize x86] AVX512/AVX/SSE intrinsic

LRY89757:
* Add the AVX512/AVX/SSE intrinsic for layernorm
* Add some test samples for elempack == 16

PR #4065, opened 2022-07-21
Layer Norm x86 SIMD Optimizations

LinHeLurking:
This PR provides some SIMD optimizations for `LayerNorm`, both for packed or unpacked tensors.


tencent-adm:
[CLA assistant check](https://cla.opensource.tencent.com/Tencent/ncnn?pullRequest=4065) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.opensource.tencent.com/Tencent/ncnn?pullRequest=4065) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.opensource.tencent.com/check/Tencent/ncnn?pullRequest=4065) it.</sub>

nihui (commented):
for simd register horizontal sum, there is utility function in x86_usability.h
for avx/fma multiply-add intrinsics, there is wrapper comp_fmadd function in x86_usability.h

use `size * elempack` as the loop count when applicable, so you can merge multiple for loop code blocks into one

LinHeLurking:
> for simd register horizontal sum, there is utility function in x86_usability.h for avx/fma multiply-add intrinsics, there is wrapper comp_fmadd function in x86_usability.h
> 
> use `size * elempack` as the loop count when applicable, so you can merge multiple for loop code blocks into one

I think I do not need SIMD register horizontal summation because the length of tensor varies. But AVX/FMA fmadd wrappers in `x86_usability.h` are useful. I've adopted them.

But I'm not sure how to merge multiple loop blocks into one by using `size * elempack` as the loop count. I think the packed layout is different with the normal one. And when handling unpacked tensors, I can calculate along dimensions. But when handling packed tensors, branches must be dispatched according to the `elempack` parameter, which is implicitly decided by CPU vector instruction set availability.

nihui:
suppose v is data from tensor, and a is the weight(such as alpha beta gamma etc.)

pack1
```
for loop1
{
    v1 + a1
}
```

pack4
```
for loop4
{
    v4 + a4
}
```

pack8
```
for loop8
{
    v8 + a8
}
```

pack16
```
for loop16
{
    v16 + a16
}
```

unified pack
```
// prepare a4 a8 a16 if pack1
a1 = a1
a4 = a1 a1 a1 a1
a8 = a4 a4
a16 = a8 a8

// prepare a8 a16 if pack4
a1 = undefined
a4 = a4
a8 = a4 a4
a16 = a8 a8

// prepare a16 if pack8
a1 = undefined
a4 = undefined
a8 = a8
a16 = a8 a8

//  if pack16
a1 = undefined
a4 = undefined
a8 = undefined
a16 = a16

for loop16
{
    v16 + a16
}
for loop8
{
    v8 + a8
}
for loop4
{
    v4 + a4
}
for loop1
{
    v1 + a1
}

```

LinHeLurking:
> suppose v is data from tensor, and a is the weight(such as alpha beta gamma etc.)
> 
> pack1
> 
> ```
> for loop1
> {
>     v1 + a1
> }
> ```
> 
> pack4
> 
> ```
> for loop4
> {
>     v4 + a4
> }
> ```
> 
> pack8
> 
> ```
> for loop8
> {
>     v8 + a8
> }
> ```
> 
> pack16
> 
> ```
> for loop16
> {
>     v16 + a16
> }
> ```
> 
> unified pack
> 
> ```
> // prepare a4 a8 a16 if pack1
> a1 = a1
> a4 = a1 a1 a1 a1
> a8 = a4 a4
> a16 = a8 a8
> 
> // prepare a8 a16 if pack4
> a1 = undefined
> a4 = a4
> a8 = a4 a4
> a16 = a8 a8
> 
> // prepare a16 if pack8
> a1 = undefined
> a4 = undefined
> a8 = a8
> a16 = a8 a8
> 
> //  if pack16
> a1 = undefined
> a4 = undefined
> a8 = undefined
> a16 = a16
> 
> for loop16
> {
>     v16 + a16
> }
> for loop8
> {
>     v8 + a8
> }
> for loop4
> {
>     v4 + a4
> }
> for loop1
> {
>     v1 + a1
> }
> ```

Thanks. Now I managed to merge many cases into one.

src/layer/x86/layernorm_x86.h
```diff
@@ -0,0 +1,22 @@
...
+class LayerNorm_x86 : virtual public LayerNorm
+{
+public:
+    LayerNorm_x86();
+
+    virtual int forward_inplace(Mat& bottom_top_blob, const Option& opt) const;
+
+protected:
+    NCNN_FORCEINLINE void fast_1d_layer_norm(float* ptr, int elempack, int elemcount, int size) const;
+    NCNN_FORCEINLINE void fast_fmadd_fmadd(float* ptr, float* a, float* b, int elempack, int elemcount, int size) const;
```
nihui, line 17: member function never inlines in practice.
declare static inline functions in cpp. (ref lstm_x86.cpp)

nihui (commented):
add copyright header for new source

src/layer/x86/layernorm_x86.h
LinHeLurking: Thanks for advising. I've fixed that.

nihui:
diff coverage is not good enough from https://app.codecov.io/gh/Tencent/ncnn/pull/4065
it may be a good idea to steal some test cases from https://github.com/Tencent/ncnn/pull/4060
add more tests to test_layernorm.cpp

LinHeLurking:
> diff coverage is not good enough from https://app.codecov.io/gh/Tencent/ncnn/pull/4065 it may be a good idea to steal some test cases from #4060 add more tests to test_layernorm.cpp

I've added some test cases about 16-packed tensors.

But I'm confused about the diff coverage. Most files shown in https://app.codecov.io/gh/Tencent/ncnn/pull/4065 are not modified or even influenced by this PR. I have no idea that how `LayerNorm_x86` affects them.

nihui:
> > diff coverage is not good enough from https://app.codecov.io/gh/Tencent/ncnn/pull/4065 it may be a good idea to steal some test cases from #4060 add more tests to test_layernorm.cpp
> 
> I've added some test cases about 16-packed tensors.
> 
> But I'm confused about the diff coverage. Most files shown in https://app.codecov.io/gh/Tencent/ncnn/pull/4065 are not modified or even influenced by this PR. I have no idea that how `LayerNorm_x86` affects them.

It often fails in that way.
Care about modified files we known.

src/layer/x86/layernorm_x86.h
```diff
@@ -0,0 +1,32 @@
+// Tencent is pleased to support the open source community by making ncnn available.
+//
+// Copyright (C) 2017 THL A29 Limited, a Tencent company. All rights reserved.
```
nihui, line 3: 2022

src/layer/x86/layernorm_x86.cpp
```diff
@@ -0,0 +1,483 @@
+// Tencent is pleased to support the open source community by making ncnn available.
+//
+// Copyright (C) 2017 THL A29 Limited, a Tencent company. All rights reserved.
```
nihui, line 3: 2022

src/layer/x86/layernorm_x86.cpp
```diff
@@ -0,0 +1,483 @@
...
+#endif // __SSE2__
+    if (elempack == 1)
+    {
+        for (; i < size; ++i, ++ptr)
+        {
+            sum += *ptr;
+        }
+        *mean = sum / elemcount;
+    }
+}
```
nihui, line 102: This may not be what I expected :P

we can make the main loop benefit from any simd isa available, no matter what the elempack is, but handle the elempack thing in post processing block.

sample code

```cpp
    int i = 0;
#if __SSE2__
#if __AVX__
#if __AVX512F__
    __m512 _sum_avx512 = _mm512_setzero_ps();
    for // step 16
    {
        // add to _sum_avx512
    }
#endif // __AVX512F__
    __m256 _sum_avx = _mm256_setzero_ps();
    for // step 8
    {
        // add to _sum_avx
    }
#endif // __AVX__
    __m128 _sum = _mm_setzero_ps();
    for // step 4
    {
        // add to _sum
    }
#endif // __SSE2__
    float sum = 0.0f;
    for // step 1
    {
        // add to sum
    }

    // ok, now we have _sum_avx512 _sum_avx _sum and sum
    // handle for different elempack
#if __SSE2__
#if __AVX__
#if __AVX512F__
    if (elempack == 16)
    {
        // write _sum_avx512 to mean
    }
#endif // __AVX512F__
    if (elempack == 8)
    {
#if __AVX512F__
        // split _sum_avx512 into two parts and add to _sum_avx
#endif // __AVX512F__
        // write _sum_avx to mean
    }
#endif // __AVX__
    if (elempack == 4)
    {
#if __AVX__
#if __AVX512F__
        // split _sum_avx512 into two parts and add to _sum_avx
#endif // __AVX512F__
        // split _sum_avx into two parts and add to _sum
#endif // __AVX__
        // write _sum to mean
    }
#endif // __SSE2__
    if (elempack == 1)
    {
#if __SSE2__
#if __AVX__
#if __AVX512F__
        // horizontal _sum_avx512 and add to sum
#endif // __AVX512F__
        // horizontal _sum_avx and add to sum
#endif // __AVX__
        // horizontal _sum and add to sum
#endif // __SSE2__
        // write sum to mean
    }
```

LinHeLurking: I've considered such plans. But writing that way makes codes as complex as the original draft (where I did not combine many branches into one).
I think as long as the ISA supports, the `elempack` would be as broad as possible. (Is that correct?) If that's true, in the first part of your sample codes where it sums no matter what the `elempack` is, packed data would only be summed in certain one for loop block and unpacked data would be summed horizontally across different blocks. I think our codes do the same thing :P. My fallback condition `(... || elempack == 1)` would ensure unpacked data also benefits from SIMD operations.

nihui: consider elempack=4 case, it may be a tensor with channel 124, which can not be perfectly divided by 16.
In the current implementation, it will fallback to __m128 block, even avx/avx512 is supported.

LinHeLurking: I see that. Newer commits cancel the relation between SIMD instruction set and `elempack`.

nihui:

PR #4075, opened 2022-07-22
RVV: BatchNorm with fp16s(a) support

thelastlin:
RVV: BatchNorm, fp16s(a)/fp32, with elempack support
(Request for comment)


nihui:

PR #4078, opened 2022-07-23
RVV: InstanceNorm with fp16s(a) support

thelastlin:
RVV: InstanceNorm, fp32/fp16, w/ elempack support


nihui:

PR #4164, opened 2022-08-21
OpenMP fault error in quick sorting recursive function

magicse:
This variant of recursive quick sorting function with OpenMP produce fault error. 
I made changes for correct working with OpenMP support enable without fault error.

### DIFF
--- a/examples/rfcn.cpp
+++ b/examples/rfcn.cpp
@@ -61,17 +61,16 @@ static void qsort_descent_inplace(std::vector<Object>& objects, int left, int ri
         }
     }
 
-    #pragma omp parallel sections
-    {
-        #pragma omp section
+   
+        #pragma omp task
         {
             if (left < j) qsort_descent_inplace(objects, left, j);
         }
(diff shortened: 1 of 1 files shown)


magicse:
Recursive calls function didn't work correct with "#pragma omp parallel sections" and produce core fault error on x86

magicse:
Essentially, what is happening is you have created an array N in shared memory, and when you do a #pragma omp parallel sections, you're spawning as many worker threads as there are #pragma omp section's. Each time a worker thread tries to access and modify elements of a, it will execute a series of instructions: "read the n'th value of N from the given address", "modify the n'th value of N", "write the n'th value of N back to the given address". Since you have multiple threads with no locking or synchronization, the read, modify, and write instructions may be executed in any order by multiple processors, so the threads may overwrite each other's modifications or read a non-updated value.
[OpenMP to parallelize QuickSort](https://stackoverflow.com/questions/16007640/openmp-parallel-quicksort)

PR #4190, opened 2022-09-07
Memory Pool Improvement For Variadic Sized Inputs

LinHeLurking:
# A Better Memory Pool With Simple Greedy Strategy

This pull request contains several modifications on current ncnn pool allocators.

## Basic Assumption

The program runs in multiple stages. In each stage, memory sizes the program is asking for are lied in an (relatively small) interval. When program goes into a new stage, the size interval will change. The intervals may or may not overlap.

## Double Ended Greedy Strategy

If all cached chunks in the pool cannot satisfy current allocation, the allocator tries to remove an outdated chunk.

- If this allocation size is larger than any previous chunk, remove the smallest chunk (see below).

```
                NEW
           ^============^
        ___^_____       ^
       |   ^     |      ^
----------------------------------------
         OLD        ^    
                    |
                    |
                ALLOCATION
```

- If this allocation size is smaller than any previous chunk, remove the largest chunk.
- If this allocation size is between the largest and smallest, do not remove.

## Simulated Results

Following codes can be used to simulate a multiple stage running scenario.

```cpp
int thrd_share_allocator_test(int repeat, ncnn::Allocator* allocator)
{
    // Fix seed for better reproducibility.
    std::mt19937 rng(0xbadf00d);
    std::uniform_int_distribution<size_t> dist;

    std::deque<ncnn::Mat> occupied;
    int batch_size = 30;
    int batch_num = repeat;
    for (int batch_id = 0; batch_id < batch_num; ++batch_id)
    {
        // Centroid and radius absolute size does not matter.
        // You can scale them arbitrarily.
        size_t centroid = std::max(1UL, dist(rng)) % 100000000;              // max centroid is 100 MB
        size_t radius = std::min(centroid, std::min(dist(rng), 30000000UL)); // max radius is 30 MB

        std::uniform_int_distribution<size_t> size_dist(centroid - radius, centroid + radius);
        std::vector<void*> allocated;
        for (int i = 0; i < batch_size; ++i)
        {
            size_t size = size_dist(rng) & ~0x13FFF; // minimal step is 5 KB
            void* ptr = allocator->fastMalloc(size);
            allocated.push_back(ptr);
        }
        std::this_thread::sleep_for(std::chrono::milliseconds(500));
        for (void* ptr : allocated)
        {
            allocator->fastFree(ptr);
        }
    }
    return 0;
}

int test_allocator(int thrd_num = 4, int repeat = int(1e1))
{
    auto allocator = new ncnn::PoolAllocator;
    std::vector<std::future<int> > futures;
    for (int i = 0; i < thrd_num; ++i)
    {
        futures.emplace_back(std::async(thrd_share_allocator_test, repeat, allocator));
    }
    int flag = 0;
    for (auto& future : futures)
    {
        flag |= future.get();
    }
    return flag;
}

int main()
{
    return 0 || test_allocator();
}
```

Memory usage:
(Memory usage is gathered with valgrind: `valgrind --tool=massif --time-unit=ms <executable>`)

image


LinHeLurking:
I'm not sure why old commits occur here. Actually only [cee9d22](https://github.com/Tencent/ncnn/pull/4190/commits/cee9d22010cc683ffecd223eb5e4c151832046bc) is the thing I did the magic.
😆

nihui:
please investigate sanitizer error

LinHeLurking:
I've tried but cannot reproduce the error of `test_squeezenet`. Is there any way to run CTest with extra checks on?

nihui:
cmake -DCMAKE_BUILD_TYPE=debug -DNCNN_ASAN=ON -DNCNN_BUILD_TESTS=ON -DNCNN_BUILD_TOOLS=OFF -DNCNN_BUILD_EXAMPLES=OFF ..

LRY89757:
> I've tried but cannot reproduce the error of `test_squeezenet`. Is there any way to run CTest with extra checks on?

May be you can refer to this file [linux-x64-cpu-gcc-san.yml](https://github.com/Tencent/ncnn/blob/master/.github/workflows/linux-x64-cpu-gcc-san.yml) about this [line](https://github.com/Tencent/ncnn/blob/d30fc825d404a8c575cd200af802fe9e51e2f729/.github/workflows/linux-x64-cpu-gcc-san.yml#L37) :)

LinHeLurking:
(First of all, sorry for late response.)

It turns out that ASAN switch of CMAKE is ignored if it is set in VS Code `setting.json` file. (But why ???)
After all, I've fixed the sanitizer error. 😆

src/allocator.cpp
```diff
@@ -33,6 +33,7 @@ class PoolAllocatorPrivate
     Mutex budgets_lock;
     Mutex payouts_lock;
     unsigned int size_compare_ratio; // 0~256
+    static const size_t size_threshold = 10;
```
nihui, line 36: `size_threshold` does not seem expressive enough
`budget_count_threshold` or `budget_drop_threshold` might be better?

How about adding a setter api that allows the user to control whether a more aggressive or conservative budget recycling strategy is required?

src/allocator.cpp
```diff
@@ -104,11 +105,20 @@ void* PoolAllocator::fastMalloc(size_t size)
     d->budgets_lock.lock();
 
     // find free budget
-    std::list<std::pair<size_t, void*> >::iterator it = d->budgets.begin();
+    std::list<std::pair<size_t, void*> >::iterator it = d->budgets.begin(), it_max = d->budgets.end(), it_min = d->budgets.end();
```
nihui, line 108: Why are these max/min iterators initialized to end() instead of begin() ?
If the budgets are empty, `if (d->budgets.size() >= d->size_threshold)` will always be false.

src/allocator.cpp
LinHeLurking: Renamed & Added a setter.

nihui:

PR #4205, opened 2022-09-17
I added one more project to the list of examples.

magicse:
I added one more project to the list of examples. Dedicated to coloring black and white photos.


nihui:

PR #4303, opened 2022-10-25
x86 unified fc fp32/fp16s

nihui:


nihui:
fp32 fall into fp16s by default, that is the issue

PR #4333, opened 2022-11-05
fix:pnnx-softmax matching any

EdVince:


nihui (approved):

nihui:

PR #4338, opened 2022-11-08
[WIP] match inplace slice copy pattern, rewrite copy uses

nihui:


nihui:
lesson learned about inplace operation

```python
import torch

class Model(nn.Module):
    def __init__(self):
        super(Model, self).__init__()

    def forward(self, x):
        x[2:5,...] += 1
        return x

def test():
    net = Model()
    net.eval()

    x = torch.zeros(8, 5)

    a = net(x)
    # a has some ones now

    b = net(x)
    # a has some two now !!!
```

PR #4356, opened 2022-11-18
Support mat.numpy() in Python

csukuangfj:
New APIs:
https://github.com/Tencent/ncnn/blob/70460dc68c57a28408cd7e6ef2b828be8db510b2/python/tests/test_mat.py#L283-L295

Motivation:
PyTorch tensors have a method called `numpy()` to convert a tensor to an numpy array.
This PR also adds one for `ncnn::Mat`.

Also, it adds support to return a numpy array of a specified type, e.g., `np.int32`. For instance, for `nn.Embedding`, it supports only `int32` tensors and this PR can let users to extract the output of an embedding layer in Python.

### DIFF
--- a/python/src/main.cpp
+++ b/python/src/main.cpp
@@ -114,857 +114,857 @@ PYBIND11_MODULE(ncnn, m)
 
     py::class_<Allocator, PyAllocator<> >(m, "Allocator");
     py::class_<PoolAllocator, Allocator, PyAllocatorOther<PoolAllocator> >(m, "PoolAllocator")
-        .def(py::init<>())
-        .def("set_size_compare_ratio", &PoolAllocator::set_size_compare_ratio, py::arg("src"))
-        .def("clear", &PoolAllocator::clear)
-        .def("fastMalloc", &PoolAllocator::fastMalloc, py::arg("size"))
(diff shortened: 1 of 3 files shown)


python/src/main.cpp
```diff
@@ -114,939 +114,857 @@ PYBIND11_MODULE(ncnn, m)
...
+             py::arg("w"), py::arg("h"), py::arg("c"), py::kw_only(),
+             py::arg("elemsize") = 4, py::arg("elempack") = 1, py::arg("allocator") = nullptr)
+        .def(py::init<int, int, int, int, size_t, int, Allocator*>(),
+             py::arg("w"), py::arg("h"), py::arg("d"), py::arg("c"), py::kw_only(),
+             py::arg("elemsize") = 4, py::arg("elempack") = 1, py::arg("allocator") = nullptr)
+
+        .def(py::init<const Mat&>(), py::arg("m"))
+
+        .def(py::init([](py::buffer const b) {
+                 py::buffer_info info = b.request();
```
csukuangfj, line 246: Only this function is changed and a new function `numpy()` is added. All other changes are formated by `clang-format` in this file.

PR #4359, opened 2022-11-19
update release ci

nihui:


nihui:
cmake find x86 python library in windows arm64 build  :(

```
        -- Found PythonInterp: C:/Users/runneradmin/AppData/Local/Temp/cibw-run-s4fba8s7/cp39-win_arm64/build/venv/Scripts/python.exe (found suitable version "3.9.10", minimum required is "3.6")
        -- Found PythonLibs: C:/Users/runneradmin/AppData/Local/pypa/cibuildwheel/Cache/nuget-cpython/python.3.9.10/tools/libs/python39.lib
```

PR #4376, opened 2022-11-28
Unpack tuple/list input

csukuangfj:
For some networks, e.g., RNN/LSTM, we need to provide some previous states as input.

Example code
```python3
#!/usr/bin/env python3

import torch
import torch.nn as nn
from typing import List


class Foo(nn.Module):
    def forward(self, x, states: List[torch.Tensor]):
        return x, states[0] + states[1]


def main():
    f = Foo()

    x = torch.rand(2, 3, 4)
    y = [x, x]
    m = torch.jit.trace(f, (x, y))
    print(m.graph)
    m.save("m.pt")


if __name__ == "__main__":
    torch.manual_seed(20221128)
    main()
```

### DIFF
--- a/tools/pnnx/src/CMakeLists.txt
+++ b/tools/pnnx/src/CMakeLists.txt
@@ -282,6 +282,7 @@ set(pnnx_pass_level3_SRCS
     pass_level3/fuse_einsum_operands.cpp
     pass_level3/fuse_expression.cpp
     pass_level3/fuse_index_expression.cpp
+    pass_level3/fuse_input_unpack.cpp
     pass_level3/fuse_multiheadattention_unpack.cpp
     pass_level3/fuse_rnn_unpack.cpp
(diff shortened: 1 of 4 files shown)


nihui:
we need level0 pass otherwise shape inference will fail

csukuangfj:
> we need level0 pass otherwise shape inference will fail

Do you mean we should move it to pass_level0 ?

nihui:
> > we need level0 pass otherwise shape inference will fail
> 
> Do you mean we should move it to pass_level0 ?

yeah, implement flatten inputs on torch graph, and apply the pass before shape inference

reference code
https://github.com/pytorch/pytorch/blob/3120054c151d37fdf43963dbb60ab420908f48cf/torch/csrc/jit/passes/lower_tuples.cpp#L185

nihui:
move to https://github.com/Tencent/ncnn/pull/4498

PR #4384, opened 2022-12-01
Merge master from tecent/ncnn

csukuangfj:
Use LSTM with projection from tecent/ncnn so that we can use `use_packing_layout = true`.

For a test wave, it reduces the decoding time from 14.33 seconds to 1.76 seconds.


csukuangfj:
不好意思， 弄错了

PR #4418, opened 2022-12-16
Flush benchmark of some CPU model by tag 20221128

wzyforgit:
Flush RTX3090、FT-2000、3A3000、3A4000、3A5000 benchmark data，add SW831 benchmark data.


nihui:

PR #4444, opened 2022-12-29
convolution deconvolution and matmul arm use sgemm

nihui:


nihui:
move to https://github.com/Tencent/ncnn/pull/4572 and https://github.com/Tencent/ncnn/pull/4594

PR #4566, opened 2023-03-11
move floor and ceil sse_function from unaryOp to sse_mathfun

Yoh-Z:


nihui:
move __SSE4_1__ block to mathfun

nihui (approved):

nihui:

PR #4570, opened 2023-03-13
Add SSE2 implementation of asin in x86 targets.

MouriNaruto:
Just do some implementations but I cannot explain the principle because I dump the implementations from compiled binaries.

I have tried my best the explain the implementation with the ChatGPT's help.

Have a good time!

Kenji Mouri


nihui (approved):

nihui:

PR #4583, opened 2023-03-22
update build doc for Centos users

inisis:
update build doc for Centos users, hope this helps


nihui:

PR #4604, opened 2023-04-05
fix typos in citation and benchmark docs

zchrissirhcz:


nihui:

PR #4620, opened 2023-04-13
Add FMA optimization for SSE2 implementation of asin, acos and atan in x86 targets.

MouriNaruto:
According to the Intel Intrinsics Guide, it looks like it will save at least 20 CPU cycles when using them.

Enjoy!

Kenji Mouri


nihui (approved):

nihui:

PR #4633, opened 2023-04-17
Add SSE and AVX implementation of atan2 in x86 targets.

MouriNaruto:
This implementation is based on SSE, AVX and AVX512F implementation of atan in x86 targets.

Note: Revert AVX512F implementation of atan2 in x86 targets because it will break “windows-x86-cpu / vs2017” CI.

Formula Reference: https://mazzo.li/posts/vectorized-atan2.html

Enjoy!

Kenji Mouri


nihui:

PR #4641, opened 2023-04-18
Add AVX512F implementation of atan2 in x86 targets.

MouriNaruto:
This implementation is based on AVX512F implementation of atan in x86 targets.

Note: Related to the implementation in https://github.com/Tencent/ncnn/pull/4633, do some workaround for MSVC 2017 for passing “windows-x86-cpu / vs2017” CI.

Formula Reference: https://mazzo.li/posts/vectorized-atan2.html

Enjoy!

Kenji Mouri


nihui:

PR #4647, opened 2023-04-19
fix fprintf argument type mismatch.

kernelbin:
return type of GetLastError is DWORD, `%d` should be used instead.


nihui (approved):

nihui:

PR #4648, opened 2023-04-19
Fix core dump in benchmark

triple-Mu:
Fix core dump in benchmark


nihui:

PR #4653, opened 2023-04-20
Fix destory in gpu

triple-Mu:
Fix destory in gpu


nihui:

PR #4659, opened 2023-04-21
move x86 abs_mathfun to x86_mathfun

Yoh-Z:


nihui:

PR #4669, opened 2023-04-26
fix typo in PNNX groupnorm error message

zchrissirhcz:
I guess this is an typo.
If it is not an typo, feel free to close it.


nihui:

PR #4691, opened 2023-05-04
Update loongson 3a4000 test score

MobtgZhang:
Update the loongson 3a4000 test scores


nihui:

PR #4693, opened 2023-05-04
I added yolov7 tiny to the model zoo since it already exists in the ncnn-assets

kennybradley:
I saw that yolov7-tiny within https://github.com/nihui/ncnn-assets/tree/master/models so it could be available to the python interface if there was a wrapper that executed the network call and post processing.

I added the yolov7.py file and updated the model_zoo and model_store so that it can transparently run within the python interface with model_zoo.

I followed the format of the other model_zoo entries as closely as I could.  

I added an optional argument use_strides to the arguments that restricts which strides to extract from the network in case users know ahead of time the size of the objects they wish to detect.  Since the anchor sizes are strongly connected to the size of the objects detected you can speed up yolov7 by targeting just the output layers that you are interested in.

### DIFF
--- a/python/ncnn/model_zoo/model_store.py
+++ b/python/ncnn/model_zoo/model_store.py
@@ -37,6 +37,8 @@ _model_sha1 = {
         ("69d128b42b70fb790e9d3ccabcf1b6e8cc2859fe", "yolov4-opt.bin"),
         ("6fa8ccc8cabc0f5633ab3c6ffa268e6042b8888f", "yolov5s.param"),
         ("0cbab3664deb090480ea748c1305f6fe850b9ac4", "yolov5s.bin"),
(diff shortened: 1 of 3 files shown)


tencent-adm:
[CLA assistant check](https://cla.opensource.tencent.com/Tencent/ncnn?pullRequest=4693) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.opensource.tencent.com/Tencent/ncnn?pullRequest=4693) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.opensource.tencent.com/check/Tencent/ncnn?pullRequest=4693) it.</sub>

nihui:

PR #4765, opened 2023-05-30
Fix pnnx conversion index out of range

Justin62628:
fix pnnx encountering index out of range at pass 5 when evauating shape expression with static input shape.


nihui:

PR #4769, opened 2023-05-31
Don't silently ignore errors in VkCompute::submit_and_wait

Upliner:
We had a problem when there was "vkQueueSubmit failed" error printed in console but ncnn::Extractor::extract doesn't return any error because return value of submit_and_wait isn't checked.

### DIFF
--- a/src/gpu.cpp
+++ b/src/gpu.cpp
@@ -1878,9 +1878,7 @@ int VulkanDevicePrivate::create_dummy_buffer_image()
     cmd.record_dummy_readonly(dummy_image_readonly);
 #endif
 
-    cmd.submit_and_wait();
-
-    return 0;
+    return cmd.submit_and_wait();
 }
 
 void VulkanDevicePrivate::destroy_dummy_buffer_image()
--- a/src/net.cpp
+++ b/src/net.cpp
@@ -162,9 +162,7 @@ int NetPrivate::upload_model()
         }
     }
 
-    cmd.submit_and_wait();
-
-    return 0;
+    return cmd.submit_and_wait();
 }
 #endif // NCNN_VULKAN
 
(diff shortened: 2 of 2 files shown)


tencent-adm:
[CLA assistant check](https://cla.opensource.tencent.com/Tencent/ncnn?pullRequest=4769) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.opensource.tencent.com/Tencent/ncnn?pullRequest=4769) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.opensource.tencent.com/check/Tencent/ncnn?pullRequest=4769) it.</sub>

src/gpu.cpp
```diff
@@ -1878,9 +1878,7 @@ int VulkanDevicePrivate::create_dummy_buffer_image()
     cmd.record_dummy_readonly(dummy_image_readonly);
 #endif
 
-    cmd.submit_and_wait();
-
-    return 0;
+    return cmd.submit_and_wait();
```
nihui, line 1881: The callee returned a possible error, but the caller did not handle it

src/net.cpp
```diff
@@ -162,9 +162,7 @@ int NetPrivate::upload_model()
         }
     }
 
-    cmd.submit_and_wait();
-
-    return 0;
+    return cmd.submit_and_wait();
```
nihui, line 165: The callee returned a possible error, but the caller did not handle it

src/net.cpp
```diff
@@ -308,9 +308,11 @@ int NetPrivate::forward_layer(int layer_index, std::vector<Mat>& blob_mats, std:
 #endif // NCNN_BENCHMARK
 
         cmd.reset();
+        if (ret != 0)
+            return ret;
     }
 
-    int ret;
+
```
nihui, line 315: remove the blank line

src/net.cpp
```diff
@@ -2499,18 +2505,18 @@ int Extractor::extract(int blob_index, Mat& feat, int type)
 #if NCNN_BENCHMARK
             cmd.create_query_pool(d->net->layers().size() * 2);
 #endif // NCNN_BENCHMARK
-
+            int ret;
```
nihui, line 2508: the ret variable is already declared earlier. this line is not necessary.

nihui (changes requested):

PR #4795, opened 2023-06-12
Update HUAWEI KunPeng 920 platform

MobtgZhang:


nihui:

PR #4845, opened 2023-07-07
VSX toolchains: check for SSE2 support

JeremyRand:
Improves compatibility with Clang 11.

Also rename NCNN_SSE* options to NCNN_VSX_SSE* to avoid conflict between x86 and POWER (went unnoticed before because x86 doesn't have an option for toggling SSE 4.1).


nihui:

PR #4860, opened 2023-07-15
c_api: expose layer_to_index

Mek101:
As per title.


nihui (approved):

nihui:

PR #4865, opened 2023-07-17
c_api return null on null layer

Mek101:
Adds a check to `ncnn_layer_create_by_typeindex` and `ncnn_layer_create_by_type` to return null if the index or name don't match any known layer.

This is mostly to move segmentation faults deriving from a wrong typeindex/name near the source callsite.


Mek101:
Seems like an error in the build system?

nihui (approved):

nihui:

PR #4895, opened 2023-07-31
add RDK X3 Module benchmark

LJoson:


nihui:

PR #4911, opened 2023-08-03
support logsoftmax and logsigmoid to ncnn

XiaBing992:


nihui:
close for https://github.com/Tencent/ncnn/pull/4925

PR #4924, opened 2023-08-07
load_param_mem pybind

JeremyRand:
Ported from https://github.com/joeyballentine/ncnn-vulkan-python/pull/1

### DIFF
--- a/python/src/main.cpp
+++ b/python/src/main.cpp
@@ -949,6 +949,7 @@ PYBIND11_MODULE(ncnn, m)
 #if NCNN_STDIO
 #if NCNN_STRING
     .def("load_param", (int (Net::*)(const char*)) & Net::load_param, py::arg("protopath"))
+    .def("load_param_mem", (int (Net::*)(const char*)) & Net::load_param_mem, py::arg("mem"))
 #endif // NCNN_STRING
     .def("load_param_bin", (int (Net::*)(const char*)) & Net::load_param_bin, py::arg("protopath"))
     .def("load_model", (int (Net::*)(const char*)) & Net::load_model, py::arg("modelpath"))


tencent-adm:
[CLA assistant check](https://cla.opensource.tencent.com/Tencent/ncnn?pullRequest=4924) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.opensource.tencent.com/Tencent/ncnn?pullRequest=4924) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.opensource.tencent.com/check/Tencent/ncnn?pullRequest=4924) it.</sub>

JeremyRand:
On further inspection, I'm not 100% sure if this approach is valid from a memory-safety perspective. Does the argument get freed before the inference is run? If so, then it seems like trying to do a Pybind for `load_param_mem` would inherently be a memory safety bug. I'm not familiar enough with the internals of ncnn or Pybind to know if this is the case; curious what you think.

JeremyRand:
Given that the point of the `load_param_mem` and `load_model_mem` functions in ncnn is to have zero copies, and since Pybind inherently makes some copies (if I understand the docs correctly?), I'm not sure this binding makes any sense. Is the correct approach to use the bindings that accept a DataReader? Let me know what you think.

JeremyRand:
Hi @nihui , can you clarify why `load_param_mem` doesn't currently have a Pybind? My guess for why is as follows:

The `load_param_mem` API is intended to be used for zero-copy purposes. It's not possible to share memory between Python and C code since they have different memory layout, so when you call a Pybind with a bytes/str argument, Pybind makes a copy of that data  before passing it to C. If you do this with a zero-copy API, ncnn winds up storing a pointer to the copy that Pybind made, and that copy will get freed as soon as that Pybind method returns. Thus, ncnn would wind up storing a pointer to freed memory, which is a memory safety bug.  Thus, the correct approach (if I have a param in Python memory, not a file) is to use the Pybind that accepts a DataReader instead.

The above makes intuitive sense to me, and also explains why I see a segfault when I add a similar Pybind for `load_model_mem` (the segfault occurs sometime after the Pybind returns). If you could confirm whether the above explanation is accurate, it would be appreciated. (If it is indeed accurate, then I'll close this PR.)

nihui:
`load_param_mem`  is fine, because the param content will always be fully parsed and stored in the members of the layer

The `load_model` api that loads weights from memory does not copy memory, but only references pointer addresses, so it needs to maintain the memory life cycle when using it. This is used to cooperate with the `ncnn2mem` tool to directly convert the bin into a C static array and store it in the static area of ​​the program, so it is designed in this way to avoid additional memory usage.


For the pull request of this load_param_mem, I think it's ok  :smile:

JeremyRand:
Thanks @nihui for clarifying. So, for `load_param_mem`, this PR's approach is OK, but for `load_model`, I should use the `DataReader` API if I want to load a model that's stored in a Python `bytes` (i.e. not stored in a file)?

nihui:

JeremyRand:
> The `load_model` api that loads weights from memory does not copy memory, but only references pointer addresses, so the caller has to maintain the memory life cycle when using it. This is usually used to cooperate with the `ncnn2mem` tool which directly converts the bin file into a C static array and stores it in the static area of ​​the program, so it is designed in this way to avoid additional memory usage.

@nihui Is this also the case in Vulkan mode, or does Vulkan mode copy the memory into Vulkan before the `load_model` function returns (thus avoiding the memory safety issue)? I'm wondering if that may explain why chaiNNer's fork of ncnn, which has a Pybind for the memory API of `load_model`, hasn't run into this issue; their fork exclusively runs in Vulkan mode, so a Vulkan/CPU difference would maybe be a plausible explanation.

PR #4925, opened 2023-08-08
Add logxxx to log comp xxx rewriter where xxx = sigmoid or softmax

lrw04:
The patch adds passes to pnnx's pass_ncnn that rewrites logsigmoid/logsoftmax nodes into compositions of log nodes and sigmoid/softmax nodes.


lrw04:
Note that this is not a templated implementation; it rewrites logsigmoid/logsoftmax nodes only and will not adapt to additions of operators in the future.

lrw04:
I prefer this over making softmax/sigmoid layers also act as logsoftmax/logsigmoid layers since that would be unnatural.

nihui:
please add test for the nn.module style ones

lrw04:
I believe they were already added in the earliest commit, at `/tools/pnnx/tests/ncnn/test_nn_{LogSigmoid.py,LogSoftmax.py}`.

nihui:

PR #4940, opened 2023-08-13
support torch.t to ncnn

XiaBing992:


nihui:

PR #4966, opened 2023-08-21
load bin in a single pass

daquexian:


nihui:

PR #4987, opened 2023-08-27
[wip] onnx coverage (as much as I can)

brightening-eyes:
hello.
I am doing onnx operator coverage for NCNN (as much as I can).


nihui:
Hi, for basic math operation such as erf, please add an op_type to binaryop/unaryop

nihui:
If you are willing to contribute more operator code, please create a separate pull request for each operator

In order to maintain typeid compatibility, the new layer is placed at the end of the list in src/CMakeLists.txt, eg `ncnn_add_layer(Erf)`

brightening-eyes:
hello,
@nihui 
after adding erf and those math stuff to unaryop, how would I add it to onnx2ncnn.cpp? I mean which part of code should I modify to make it compatible?
because when a model gets converted, it complains that it can not find erf, shape, etc.

nihui:
> hello, @nihui after adding erf and those math stuff to unaryop, how would I add it to onnx2ncnn.cpp? I mean which part of code should I modify to make it compatible? because when a model gets converted, it complains that it can not find erf, shape, etc.

It should be like sin
- add op judgment
- convert type
- convert parameters

https://github.com/Tencent/ncnn/blob/0829fa49d3d74e9e8d6b053cc0bef60a06c60bce/tools/onnx/onnx2ncnn.cpp#L3897

https://github.com/Tencent/ncnn/blob/0829fa49d3d74e9e8d6b053cc0bef60a06c60bce/tools/onnx/onnx2ncnn.cpp#L5742

brightening-eyes:
hello.
I got it.
will try my best to implement as much as I can.
I will close this as it might not needed.
@nihui 
thanks for your help.

PR #4988, opened 2023-08-28
Fix potential overflow in determining the datatype

lrw04:
Counterexample: flag = 0x01ff01ff


nihui (approved):

nihui:

PR #5012, opened 2023-09-06
erf implementation

brightening-eyes:
hello.
added erf with its tests and vulkan support.


nihui:

PR #5017, opened 2023-09-09
selu and shufflechannel on x86

wnqn1597:


nihui:

PR #5018, opened 2023-09-09
celu activation

brightening-eyes:
CeLU is one of the operators in onnx, and another activation like relu, prelu, elu and so on.


nihui:
conflicts resolved for https://github.com/Tencent/ncnn/pull/5019 applied

nihui:

PR #5019, opened 2023-09-09
Celu layer and export to ncnn

wnqn1597:


nihui:

PR #5021, opened 2023-09-10
Size layer

brightening-eyes:
this is Size layer, equivalent to size operator in onnx


nihui:
tools/onnx/onnx2ncnn.cpp.orig accidently committed ?

brightening-eyes:
done.
should I add it to list of ops?

nihui:
> done. should I add it to list of ops?

yeah, add to operators doc

brightening-eyes:
done as well.

src/layer/size.cpp
```diff
@@ -0,0 +1,36 @@
...
+    support_inplace = false;
+}
+
+int Size::forward(const Mat& bottom_blob, Mat& top_blob, const Option& opt) const
+{
+    top_blob.create(1);
+    if (top_blob.empty())
+        return -100;
+
+    top_blob[0] = bottom_blob.total();
```
nihui, line 31: Due to potential channel gap, total() may be larger than the actual number.

This is used to ensure that underlying memory accesses are aligned for improved efficiency.

Example
dimension rank is 3 and w=1 h=1 c=5
Mat::total() will return align(w*h, 4) * 5 = 20

Access members such as dims, w, h, d, c and calculate it.

nihui (changes requested):

src/layer/size.cpp
```diff
@@ -0,0 +1,46 @@
...
+
+    int size;
+    if (bottom_blob.dims == 1)
+        size = bottom_blob.w;
+    else if (bottom_blob.dims == 2)
+        size = bottom_blob.w * bottom_blob.h;
+    else if (bottom_blob.dims == 3)
+        size = bottom_blob.w * bottom_blob.h * bottom_blob.c;
+    else if (bottom_blob.dims == 4)
+        size = bottom_blob.w * bottom_blob.h * bottom_blob.d * bottom_blob.c;
```
nihui, line 39: this single line is enough
`int size = bottom_blob.w * bottom_blob.h * bottom_blob.d * bottom_blob.c;`

src/layer/size.cpp
```diff
@@ -0,0 +1,36 @@
...
+
+Size::Size()
+{
+    one_blob_only = true;
+    support_inplace = false;
+}
+
+int Size::forward(const Mat& bottom_blob, Mat& top_blob, const Option& opt) const
+{
+    top_blob.create(1);
```
nihui, line 27: create with opt.blob_allocator argument

tests/test_size.cpp
```diff
@@ -0,0 +1,65 @@
...
+    int ret = test_layer<ncnn::Size>("Size", pd, weights, a);
+    if (ret != 0)
+    {
+        fprintf(stderr, "test_size failed a.dims=%d a=(%d %d %d)\n", a.dims, a.w, a.h, a.c);
+    }
+
+    return ret;
+}
+
+static int test_size_0()
```
nihui, line 33: add test for 4-d blobs

tests/test_size.cpp
```diff
@@ -0,0 +1,74 @@
...
+static int test_size(const ncnn::Mat& a)
+{
+    ncnn::ParamDict pd;
+
+    std::vector<ncnn::Mat> weights(0);
+
+    int ret = test_layer<ncnn::Size>("Size", pd, weights, a);
+    if (ret != 0)
+    {
+        fprintf(stderr, "test_size failed a.dims=%d a=(%d %d %d)\n", a.dims, a.w, a.h, a.c);
```
nihui, line 27:
```suggestion
        fprintf(stderr, "test_size failed a.dims=%d a=(%d %d %d %d)\n", a.dims, a.w, a.h, a.d, a.c);
```

nihui (approved):

src/layer/size.cpp
```diff
@@ -0,0 +1,36 @@
...
+    support_inplace = false;
+}
+
+int Size::forward(const Mat& bottom_blob, Mat& top_blob, const Option& opt) const
+{
+    top_blob.create(1, 4u, opt.blob_allocator);
+    if (top_blob.empty())
+        return -100;
+
+    top_blob[0] = bottom_blob.w * bottom_blob.h * bottom_blob.d * bottom_blob.c;
```
nihui, line 31: onnx size outputs an integer tensor
The code implemented here by ncnn size will return a float tensor
For very large numbers, there is a loss of precision when using float
Will this cause difficulty in subsequent layer processing? For example, perform other addition, subtraction or modulo division operations on the result of size.

brightening-eyes: for things like rms norm etc, it might have a little impact. but, the issue is that the size of batches (which ncnn does not have by itself), will have this. other than devisions for normalizations, I dont think it has other impacts. but if necessary (which is better), I might change the types from float to int. should anything else be done to make it compatible with integers?

nihui: I'm worried that your original intention was to return integer, but you accidentally returned float tensor. Now it seems that I have unnecessary worries.  :D

Regarding batch, the process of converting the model to ncnn will require that the batch is always 1. There may be special cases, but that will also be solved in the converter, such as intepret nchw to cdhw in ncnn as a workaround. The layer itself doesn't care about it

brightening-eyes:
how would I assign it as an integer? the types are all ints, and multiplications will return an integer.
ncnn::Mat has an operator t* which can accept ints as well!. isn't it?
I assume as well that we should have integers instead of floats because size always returns an integer scalar.
your comment also brought to my attention that we need to consider having integer as the output size.

nihui:
> @nihui what about doing memcpy on output tensor rather than assigning it to its first element? does it make it better?

```cpp
// access by channel / depth
int* p = mat.channel(i);
int* p = mat.depth(i);

// access by row
int* p = mat.row<int>(i);

// access by element
int* p = mat;

// set value at 1,2,3,4 to 567
mat.channel(1).depth(2).row<int>(3)[4] = 567;
```

PR #5030, opened 2023-09-14
isNaN implementation

brightening-eyes:


Baiyuetribe:
@brightening-eyes Can you try `Gather` in implementation? This one is in many onnx models.

brightening-eyes:
@Baiyuetribe 
I will do that afterwards.
for example, pytorch uses gather when it converts the models to onnx, as onnx does not have anything as embedding. rather, it uses gather to index from a matrix of weights.
for positional embeddings (for absolutes and sometimes relatives), it uses range along with gather which both of them should be implemented.

nihui:
shall be placed in unaryop

brightening-eyes:
the thing is, for isNaN, this might seem true.
but isInf has detect_negative and detect_positive attributes which makes it completely different from unaryop.
so, in this case, it is best to avoid modifying unaryop.

PR #5042, opened 2023-09-19
sign operator

brightening-eyes:
this is one of the simplest things in onnx which I thaught to implement on ncnn.
the same is true with softsign and so on.
I might implement the vulkan version as well.


nihui:
sign should be placed in unaryop

brightening-eyes:
I'm doing it like onnxruntime.
I suppose its easier to implement and maintain than unaryop. especially when I do not have experience with loongarch stuff.

PR #5049, opened 2023-09-21
test conv1d update

FhqTreap:
in my tests, b1 may become an unpredictable value. As a bias term, I think it should be either 0 or 1.


nihui:

PR #5070, opened 2023-10-07
benchmark: add RTX A3000 6G benchmark

chainsx:


nihui:

PR #5074, opened 2023-10-09
pnnx support torch-2.1

nihui:


nihui:
additional linker flag for libtorchvision.so
```
-Wl,--no-as-needed
```

PR #5103, opened 2023-10-25
slice with indices

nihui:


csukuangfj:
太棒啦！

x825196964:
is index_ptr stores index of start or end?

PR #5144, opened 2023-11-11
Improve cmake toolchain files

zchrissirhcz:
A cleanup of `toolchains` directory's cmake files:
- Use `set` instead of `SET `
- Specify `CMAKE_FIND_ROOT_PATH_MODE_PACKAGE`'s value
- Provide default values for `CMAKE_FIND_ROOT_PATH_MODE_{PROGRAM,LIBRARY,INCLUDE,PACKAGE}`, allow user to overwrite them in cmake command args
- Indent with 2 spaces in affected files

Related reference: https://github.com/android/ndk/issues/517

The PR is divided into 4 commits but can be squashed into one if is agreed to be merged.


zchrissirhcz:
force pushed due to submodules mis-added in previous push.

nihui (changes requested):
Use 4 spaces for indentation

nihui (approved):

nihui:

PR #5149, opened 2023-11-14
Prelu layer uses  sse instruction _mm_load_ps but data can be misaligned so it must use _mm_loadu_ps

AlOa:
In the prelu_x86.cpp for loading the data  for sse instructions it is used _mm_load_ps but data can be misaligned so it must be used _mm_loadu_ps


nihui:
Hi
This problem occurs when you load the model from memory, slope_data is not guaranteed to be aligned, which will lead to misalignment

However, I found that there is another place where _mm_load_ps is used incorrectly. Please fix it together.

https://github.com/Tencent/ncnn/blob/837bbbfd5580ab4dbfc0dcee55c1d861c87784a4/src/layer/x86/prelu_x86.cpp#L84

nihui (approved):

nihui:

PR #5160, opened 2023-11-21
Update RPi 5 GPU Benchmark Result

FantasyGmm:
use mesa-23.3-rc4 driver


nihui:

PR #5178, opened 2023-11-27
Update Vulkan dependency docs

JeremyRand:
Vulkan deps are optional. `vulkan-utils` is replaced with `vulkan-tools` since Debian 10.


nihui:
Yeah
With simplevk enabled by default in the next release, the vulkan sdk is no longer needed. We will most likely need to update the documentation  :D

nihui:

PR #5213, opened 2023-12-10
add Jetson TX2 benchmark

FantasyGmm:


nihui:

PR #5245, opened 2023-12-22
requantize: Use activation_ss in fused_activation.h

MollySophia:
This fixes int8 requantization on risc-v, which has some unique precision issues.


nihui:

PR #5303, opened 2024-01-22
RVV: Refine riscv gemm fp32

Xinyu302:
Use RVV intrinsic vsseg4e32_v_f32m1, vsseg2e32_v_f32m1 to do load interleaved.
Modify transpose.


nihui:

PR #5310, opened 2024-01-25
convolution_x86: Fix typo in logging

strongtz:
This might be important as I've been urged by someone to do this. :)


nihui:

PR #5320, opened 2024-01-29
python pnnx add option to change fp16 parameter

MollySophia:


nihui (commented):
add this parameter to export()

nihui:

PR #5342, opened 2024-02-22
fix dual issue  document

luqiang-guo:
- Modified the dual issueof a55


nihui:

PR #5353, opened 2024-02-29
Fix `destroy_gpu_instance` crash

shatyuka:
`destroy_gpu_instance` crashes when `create_gpu_instance` execution fails.
for example it will crash in the `__ncnn_vulkan_instance_holder` destructor on program exit if there is no vulkan driver found.

### DIFF
--- a/src/gpu.cpp
+++ b/src/gpu.cpp
@@ -51,6 +51,7 @@ public:
     {
         instance = 0;
         created = 0;
+        glslang_initialized = false;
 
 #if NCNN_VULKAN_LOADER
         libvulkan = 0;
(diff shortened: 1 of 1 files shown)


src/gpu.cpp
```diff
@@ -2091,7 +2093,8 @@ void destroy_gpu_instance()
 
     // NCNN_LOGE("destroy_gpu_instance");
 
-    glslang::FinalizeProcess();
+    if (g_instance.glslang_initialized)
+        glslang::FinalizeProcess();
```
nihui, line 2097: set glslang_initialized to zero

shatyuka: alright forgot it

nihui (approved):

nihui:

PR #5359, opened 2024-03-02
Add implementation of build for protobuf>=22.0 on Windows

Galasnow:
And fix a missing word.


nihui:

PR #5370, opened 2024-03-11
Support LLVM OpenMP runtime for MSVC

shatyuka:
with `/openmp:llvm` compile option


nihui (approved):

nihui:

PR #5436, opened 2024-04-18
c_api: Fix function prototypes with no argument

quink-black:
Here is a big difference between C and C++. foo() in C means that the funtion takes an unspecified number of arguments, while foo(void) means the function takes no argument.

Fix -Wstrict-prototypes warning.

### DIFF
--- a/src/c_api.h
+++ b/src/c_api.h
@@ -26,7 +26,7 @@
 extern "C" {
 #endif
 
-NCNN_EXPORT const char* ncnn_version();
+NCNN_EXPORT const char* ncnn_version(void);
 
 /* allocator api */
 typedef struct __ncnn_allocator_t* ncnn_allocator_t;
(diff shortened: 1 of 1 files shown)


tencent-adm:
[CLA assistant check](https://cla.opensource.tencent.com/Tencent/ncnn?pullRequest=5436) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.opensource.tencent.com/Tencent/ncnn?pullRequest=5436) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.opensource.tencent.com/check/Tencent/ncnn?pullRequest=5436) it.</sub>

nihui (approved):

nihui:

PR #5464, opened 2024-05-17
fix mingw64 avx crash and termux build issue

TianZerL:
There is a bug for MinGW64,  witch does not perform 32-byte stack alignment when it should, so avx and avx 512 may crash if we use `__m256` or `__m512` as function argument or return value. see [here](https://stackoverflow.com/questions/30926241/wrapper-for-m256-producing-segmentation-fault-with-constructor-windows-64).

This PR fixed them by passing  `__m256` or `__m512` as const reference, and for return value, we use `NCNN_FORCEINLINE` to make sure they will be a inline function.

ncnn build in Termux will enable Android platform API by default, but the related lib will be linked only if we are using NDK, this PR make fixed it.


nihui:

PR #5471, opened 2024-05-23
fix issue: no member named create_gpu_instance in namespace ncnn

zhouwg:
This PR is intent to fix following issue:

Screenshot from 2024-05-23 20-35-46

This issue could be reproduced very easily with official release pkg [ncnn-20240410-android-vulkan.zip](https://github.com/Tencent/ncnn/releases/download/20240410/ncnn-20240410-android-vulkan.zip).

the root-cause is that NCNN_VULKAN was defined/hard-coded by CMake build script <b>and </b> many NCNN's internal codes heavily depend on the pre-defined/hard-coded NCNN_VULKAN, <b> but </b> some NCNN GPU APIs(such as create_gpu_instance, destroy_gpu_instance,  get_gpu_count...) would be called in user's code even the scenario is pure CPU inference.


tencent-adm:
[CLA assistant check](https://cla.opensource.tencent.com/Tencent/ncnn?pullRequest=5471) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.opensource.tencent.com/Tencent/ncnn?pullRequest=5471) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.opensource.tencent.com/check/Tencent/ncnn?pullRequest=5471) it.</sub>

nihui (changes requested):

PR #5503, opened 2024-06-11
RFC: Add a perf class to measure performance

quink-black:
Use Linux perf API on Linux and Android to measure cpu circles taking by a code block, fallback to get_current_time() otherwise.

It meant to compare the performance between different implementations, or the performance before and after optimization.

### DIFF
--- a/benchmark/benchncnn.cpp
+++ b/benchmark/benchncnn.cpp
@@ -81,7 +81,7 @@ private:
     std::vector<ncnn::Option> opts;
     ncnn::Perf perf;
 
-    const char *prev_comment;
+    const char* prev_comment;
     double prev_time_avg;
     int64_t prev_cpu_avg;
 };
@@ -89,15 +89,15 @@ private:
 Benchmark::Benchmark()
     :
 #if NCNN_VULKAN
-      vkdev(NULL),
-      blob_vkallocator(NULL),
-      staging_vkallocator(NULL),
+    vkdev(NULL),
+    blob_vkallocator(NULL),
+    staging_vkallocator(NULL),
 #endif
-      warmup_loop_count(8),
-      loop_count(4),
-      enable_cooling_down(true),
-      prev_comment(NULL),
-      prev_time_avg(0)
+    warmup_loop_count(8),
+    loop_count(4),
+    enable_cooling_down(true),
+    prev_comment(NULL),
+    prev_time_avg(0)
 {
     if (perf.init())
     {
--- a/benchmark/benchncnn.cpp
+++ b/benchmark/benchncnn.cpp
@@ -70,7 +70,17 @@ public:
 private:
     void run(const char* comment, const std::vector<ncnn::Mat>& _in, const ncnn::Option& opt, bool fixed_path);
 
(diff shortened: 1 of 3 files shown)


quink-black:
Usage example:

```
diff --git a/benchmark/benchncnn.cpp b/benchmark/benchncnn.cpp
index 067415a2..532ab417 100644
--- a/benchmark/benchncnn.cpp
+++ b/benchmark/benchncnn.cpp
@@ -50,6 +50,7 @@ static bool g_enable_cooling_down = true;
 
 static ncnn::UnlockedPoolAllocator g_blob_pool_allocator;
 static ncnn::PoolAllocator g_workspace_pool_allocator;
+static ncnn::Perf g_perf;
 
 #if NCNN_VULKAN
 static ncnn::VulkanDevice* g_vkdev = 0;
@@ -140,13 +141,13 @@ void benchmark(const char* comment, const std::vector<ncnn::Mat>& _in, const ncn
         }
     }
 
-    double time_min = DBL_MAX;
-    double time_max = -DBL_MAX;
-    double time_avg = 0;
+    uint64_t time_min = UINT64_MAX;
+    uint64_t time_max = 0;
+    uint64_t time_avg = 0;
 
     for (int i = 0; i < g_loop_count; i++)
     {
-        double start = ncnn::get_current_time();
+        uint64_t start = g_perf.start_timer();
         {
             ncnn::Extractor ex = net.create_extractor();
             for (size_t j = 0; j < input_names.size(); ++j)
@@ -162,9 +163,7 @@ void benchmark(const char* comment, const std::vector<ncnn::Mat>& _in, const ncn
             }
         }
 
-        double end = ncnn::get_current_time();
-
-        double time = end - start;
+        uint64_t time = g_perf.stop_timer(start);
 
         time_min = std::min(time_min, time);
         time_max = std::max(time_max, time);
@@ -173,7 +172,7 @@ void benchmark(const char* comment, const std::vector<ncnn::Mat>& _in, const ncn
 
     time_avg /= g_loop_count;
 
-    fprintf(stderr, "%20s  min = %7.2f  max = %7.2f  avg = %7.2f\n", comment, time_min, time_max, time_avg);
+    fprintf(stderr, "%20s  min = %7.2f  max = %7.2f  avg = %7.2f\n", comment, time_min / 1000.0, time_max / 1000.0, time_avg / 1000.0);
 }
 
 void benchmark(const char* comment, const ncnn::Mat& _in, const ncnn::Option& opt, bool fixed_path = true)
@@ -327,6 +326,11 @@ int main(int argc, char** argv)
         return -1;
     }
 
+    if (g_perf.init()) {
+        fprintf(stderr, "init perf failed\n");
+        return -1;
+    }
+
 #ifdef __EMSCRIPTEN__
     EM_ASM(
         FS.mkdir('/working');
```

However, the print isn't beautiful in this usecase, as the unit of stop_timer is unspecified. As the comment says, it's meant to measure relative speed than absolute value.

```
          squeezenet  min = 12429.34  max = 12640.46  avg = 12557.65
          squeezenet_int8  min = 9612.34  max = 9807.91  avg = 9707.43
          mobilenet  min = 22409.47  max = 23029.78  avg = 22710.12
```

PR #5551, opened 2024-07-01
add OrangePicm4

py1066:
doc: https://zhuanlan.zhihu.com/p/706371699


nihui:

PR #5603, opened 2024-07-25
arm/convolution_3x3_pack1to8_fp16s: prefer ldr/str over ld1/st1

quink-black:
Depending on the arch, ldr/str can be faster than ld1/st1, especially for loading to one lane form. For example

On Cortex A75,
1. execution latency of 'ldr q0' and 'ldr h0' are 5
2. execution latency of 'ld1 {v0.16b}' is 6
3. execution latency of 'ld1 {v0.h}[0]' is 8

On Cortex X3,
1. execution latency of 'ldr q0' and 'ldr h0' are 6
2. execution latency of 'ld1 {v0.16b}' is 6
3. execution latency of 'ld1 {v0.h}[0]' is 8

### DIFF
--- a/src/layer/arm/convolution_3x3_pack1to8_fp16s.h
+++ b/src/layer/arm/convolution_3x3_pack1to8_fp16s.h
@@ -68,8 +68,8 @@ static void conv3x3s1_pack1to8_fp16sa_neon(const Mat& bottom_blob, Mat& top_blob
                         "sub    %0, %0, #64                 \n"
 
                         "prfm   pldl1keep, [%1, #128]       \n"
-                        "ld1    {v0.8h}, [%1], #16          \n" // r0
-                        "ld1    {v1.4h}, [%1]               \n"
+                        "ldr    q0, [%1], #16               \n" // r0
+                        "ldr    s1, [%1]                    \n"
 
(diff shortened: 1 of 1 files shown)


src/layer/arm/convolution_3x3_pack1to8_fp16s.h
```diff
@@ -68,8 +68,8 @@ static void conv3x3s1_pack1to8_fp16sa_neon(const Mat& bottom_blob, Mat& top_blob
                         "sub    %0, %0, #64                 \n"
 
                         "prfm   pldl1keep, [%1, #128]       \n"
-                        "ld1    {v0.8h}, [%1], #16          \n" // r0
-                        "ld1    {v1.4h}, [%1]               \n"
+                        "ldr    q0, [%1], #16               \n" // r0
+                        "ldr    s1, [%1]                    \n"
```
nihui, line 72: v1.4h is 64bit, but s1 is 32bit ? use d1 ?

quink-black: Only two lanes (`v1.h[0]` and `v1.h[0`]`) are used here, isn't is?

nihui:

PR #5735, opened 2024-10-14
Fix BOF in the last channel handling logic of SSE ShuffleChannel 

### LINKED ISSUE #5734: The last channel handling logic in the SSE version of ShuffleChannel cause buffer-overflow
## context
In SSE version of ShuffleChannel, the last channel is shuffled from the offset with the half of the granularity.
Here, it causes buffer-overflow at the load of the ptr1 in the last loop.

For example, in the case of (elempack == 4) && (_group == 2 && channels % _group != 0) in [AVX512 optimization](https://github.com/Tencent/ncnn/blob/master/src/layer/x86/shufflechannel_x86.cpp#L602), 

The ptr1 initially can be accessed for the range [ptr1, ptr1+4*size), 
and the range is reduced into [ptr1, ptr1+4*size-2) after `ptr1 += 2;`. 
However, in the last for loop, it accesses [ptr1+4*size, ptr1+4*(size+1)), which leads to buffer-overflow.

Since it cause both buffer-overflow read (ptr1) and buffer-overflow write (outptr), it could lead to incorrect result of the model.
```
{
      const float* ptr0 = bottom_blob.channel(channels_per_group);
      const float* ptr1 = bottom_blob.channel(channels_per_group * 2);
      float* outptr = top_blob.channel(channels_per_group * 2);

      ptr1 += 2;

      for (int i = 0; i < size; i++)
      {
          __m128 _p0 = _mm_loadu_ps(ptr0);
          __m128 _p1 = _mm_loadu_ps(ptr1);

          __m128 _lo = _mm_unpacklo_ps(_p0, _p1);

          _mm_storeu_ps(outptr, _lo);

          ptr0 += 4;
          ptr1 += 4;
          outptr += 4;
      }
  }
```

**x86**
https://github.com/Tencent/ncnn/blob/master/src/layer/x86/shufflechannel_x86.cpp#L117
https://github.com/Tencent/ncnn/blob/master/src/layer/x86/shufflechannel_x86.cpp#L373
https://github.com/Tencent/ncnn/blob/master/src/layer/x86/shufflechannel_x86.cpp#L608

**arm**
https://github.com/Tencent/ncnn/blob/master/src/layer/arm/shufflechannel_arm.cpp#L118
https://github.com/Tencent/ncnn/blob/master/src/layer/arm/shufflechannel_arm.cpp#L365
https://github.com/Tencent/ncnn/blob/master/src/layer/arm/shufflechannel_arm.cpp#L599

## how to reproduce
1. Build with SSE in x86 or arm
2. ./test_shufflechannel

## more 
I will open a PR of the patch for this:)

junwha:
Fix https://github.com/Tencent/ncnn/issues/5734

- apply SSE logic until size-1
- apply linear logic for the last n elements


tencent-adm:
[CLA assistant check](https://cla.opensource.tencent.com/Tencent/ncnn?pullRequest=5735) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you all sign our [Contributor License Agreement](https://cla.opensource.tencent.com/Tencent/ncnn?pullRequest=5735) before we can accept your contribution.<br/>**0** out of **2** committers have signed the CLA.<br/><br/>:x: junwha<br/>:x: nihui<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.opensource.tencent.com/check/Tencent/ncnn?pullRequest=5735) it.</sub>

src/layer/x86/shufflechannel_x86.cpp
```diff
@@ -387,6 +402,21 @@ int ShuffleChannel_x86::forward(const Mat& bottom_blob, Mat& top_blob, const Opt
...
+                        *outptr = *ptr1;
+                        ptr1 += 1;
+                    }
+                    else
+                    {
+                        *outptr = *ptr0;
+                        ptr0 += 1;
+                    }
+                    outptr += 1;
+                }
```
nihui, line 419: I don't think there is a risk of out of range reading in some of the pack8/pack16 codes.

src/layer/arm/shufflechannel_arm.cpp
```diff
@@ -611,6 +641,21 @@ int ShuffleChannel_arm::forward_bf16s_fp16s(const Mat& bottom_blob, Mat& top_blo
...
+                        *outptr0 = *ptr1;
+                        ptr1 += 1;
+                    }
+                    else
+                    {
+                        *outptr0 = *ptr0;
+                        ptr0 += 1;
+                    }
+                    outptr0 += 1;
+                }
```
nihui, line 658: unroll this

src/layer/x86/shufflechannel_x86.cpp
```diff
@@ -134,6 +134,21 @@ int ShuffleChannel_x86::forward(const Mat& bottom_blob, Mat& top_blob, const Opt
...
```
nihui, line 151: I don't think there is a risk of out of range reading in some of the pack8/pack16 codes.

src/layer/arm/shufflechannel_arm.cpp
```diff
@@ -378,6 +393,21 @@ int ShuffleChannel_arm::forward_bf16s_fp16s(const Mat& bottom_blob, Mat& top_blo
...
```
nihui, line 410: I don't think there is a risk of out of range reading in some of the arm pack8 codes.

src/layer/arm/shufflechannel_arm.cpp
```diff
@@ -130,6 +130,21 @@ int ShuffleChannel_arm::forward(const Mat& bottom_blob, Mat& top_blob, const Opt
...
```
nihui, line 147: unroll this

nihui (changes requested):

junwha:
Hi @nihui,
Thank you for the fast response and review!
I will reflect the unrolling.

Could I ask why do you think the out-of-bound risk does not exist in pack8/16 codes?
As far as I think, it also accesses (start of the channel) + (size-1)*(element size) + (element size)/2 in the same manner, which is beyond the size of the channel.

Thank you for your effort on this issue!

nihui (approved):

nihui:

PR #5806, opened 2024-11-28
Update windows-arm.yml

Shironana817:


nihui:
move to https://github.com/Tencent/ncnn/pull/5811

PR #5811, opened 2024-12-03
remove support for arm32

Shironana817:


nihui:

PR #5927, opened 2025-03-04
Support better API to detect big little core in windows after win7

futz12:
Support better API to detect big little core in windows after win7

By the way,I add license in test_cpu.cpp

### DIFF
--- a/src/cpu.cpp
+++ b/src/cpu.cpp
@@ -1577,7 +1577,7 @@ static void initialize_cpu_thread_affinity_mask(ncnn::CpuSet& mask_all, ncnn::Cp
         glpie(RelationProcessorCore, nullptr, &bufferSize);
         std::vector<BYTE> buffer(bufferSize);
         if (!GetLogicalProcessorInformationEx(RelationProcessorCore,
-            (SYSTEM_LOGICAL_PROCESSOR_INFORMATION_EX*)(buffer.data()), &bufferSize))
+                                              (SYSTEM_LOGICAL_PROCESSOR_INFORMATION_EX*)(buffer.data()), &bufferSize))
         {
             NCNN_LOGE("GetLogicalProcessorInformationEx failed");
             return;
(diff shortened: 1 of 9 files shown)


tencent-adm:
[CLA assistant check](https://cla.opensource.tencent.com/Tencent/ncnn?pullRequest=5927) <br/>All committers have signed the CLA.

src/cpu.cpp
```diff
@@ -36,6 +36,7 @@
 #ifdef _MSC_VER
 #include <intrin.h>    // __cpuid()
 #include <immintrin.h> // _xgetbv()
+#include <unordered_map>
```
nihui, line 39: unordered_map is not available in c++03 / simplestl atm. use vector instead

nihui (changes requested):

src/cpu.cpp
futz12: i replace it with map

src/cpu.cpp
```diff
@@ -1557,45 +1557,145 @@ static void initialize_cpu_thread_affinity_mask(ncnn::CpuSet& mask_all, ncnn::Cp
...
+            mask_little.disable_all();
+            mask_big = mask_all;
+        }
         else
-            mask_big.enable(i);
+        {
+            for (int i = 0; i < g_cpucount; i++)
+            {
+                bool isECore = false;
+                for (auto& p : processorCoreType)
```
nihui, line 1635: ranged for

tests/test_cpu.cpp
```diff
@@ -117,11 +117,33 @@ static int test_cpu_powersave()
...
 static int test_cpu_info()
 {
+    int cpucount = ncnn::get_cpu_count();
+    int bigcpucount = ncnn::get_big_cpu_count();
+    int littlecpucount = ncnn::get_little_cpu_count();
+
+    fprintf(stderr, "cpucount = %d\n", cpucount);
+    fprintf(stderr, "bigcpucount = %d\n", bigcpucount);
+    fprintf(stderr, "littlecpucount = %d\n", littlecpucount);
+
```
nihui, line 133: test bigcpucount + littlecpucount == cpucount
test bigcpucount <= cpucount
test littlecpucount < cpucount

return -1 on error

nihui (approved):

nihui:

PR #6102, opened 2025-06-02
workaround for device lost error on windows + nvidia graphics, use 4k offset alignment for all buffer allocation

nihui:


nihui:
close for https://github.com/Tencent/ncnn/pull/6296

PR #6101, opened 2025-06-01
Switch to putenv to fix compilation with llvm-mingw

zhuzeitou:
Replaces `setenv` with `putenv` because `setenv` is not available on some toolchains (e.g., llvm-mingw, clang-cl), leading to compilation errors.

PR #6147, opened 2025-06-21
Use platform-specific APIs for environment variables

zhuzeitou:
The previous patch (https://github.com/Tencent/ncnn/pull/6101) used `putenv` as a quick fix for Windows compatibility. However, `putenv` is a legacy API and not the recommended choice.

This commit replaces the single `putenv` call with the most appropriate function for each platform:
- On Windows, it now uses the modern and secure `_putenv_s`.
- On Unix-like systems, it uses the standard `setenv`.


nihui (approved):

nihui:

PR #6165, opened 2025-07-10
feat: add cmake print ver support

Willaaaaaaa:
# feat: add cmake print ver support(#6139 )

## Main changes
- Modified `./cmake/ncnnConfig.cmake.in` to print `ncnn` version to the console when `find_package(ncnn)` is called.
- Follow the `QUIET` argument of the `find_package` command. When using `QUIET`, users will not get the version message.

> There are multiple version variables in `./CMakeLists.txt` and I'm not sure if just printing the date information is enough.

## 主要修改
- 修改了 `cmake/ncnnConfig.cmake.in` 文件，使得当 `find_package(ncnn)` 成功执行时，会在 console 打印 ncnn 的版本信息。
- 遵循 `find_package` 命令的 `QUIET` 参数，在静默模式下不输出版本信息。

> 在 `./CMakeLists.txt` 中有多个关于版本的变量，我不确定只打印这个日期信息是否是足够的。

image

### DIFF
--- a/cmake/ncnnConfig.cmake.in
+++ b/cmake/ncnnConfig.cmake.in
@@ -1,3 +1,4 @@
+set(NCNN_VERSION @NCNN_VERSION@)
 set(NCNN_OPENMP @NCNN_OPENMP@)
 set(NCNN_THREADS @NCNN_THREADS@)
 set(NCNN_VULKAN @NCNN_VULKAN@)
@@ -44,3 +45,10 @@ if(NCNN_VULKAN)
 endif()
 
 include(${CMAKE_CURRENT_LIST_DIR}/ncnn.cmake)
+
+if(TARGET ncnn)
+    set(ncnn_FOUND TRUE)
+    if(NOT ncnn_FIND_QUIETLY)
+        message(STATUS "Found ncnn: ${NCNN_VERSION}")
+    endif()
+endif()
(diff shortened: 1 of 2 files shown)


tencent-adm:
[CLA assistant check](https://cla.opensource.tencent.com/Tencent/ncnn?pullRequest=6165) <br/>All committers have signed the CLA.

Willaaaaaaa:
Moreover, I have noticed that the existing `cmake.in` file uses `find_package(sth)` to add dependencies, but it seems that we can use `find_dependency` to handle it. The advantage of this is that the `REQUIRED` and `QUIET` parameters can be passed down.

nihui:
https://stackoverflow.com/questions/51699512/how-does-cmakes-export-know-what-version-the-target-is

use `write_basic_package_version_file`

https://cmake.org/cmake/help/v3.0/module/CMakePackageConfigHelpers.html#example-generating-package-files

let `find_package(ncnn VERSION 20250503)` work

Willaaaaaaa:
According to [cmake find_package doc](https://cmake.org/cmake/help/latest/command/find_package.html#typical-usage), unlike `project(Lib VERSION 0.1)`, `find_package(ncnn 20250503)` without the `VERSION` keyword seems to be the right syntax.

nihui (approved):

nihui:

PR #6174, opened 2025-07-11
Skip int8 model on GPU and merge from upstream

c8s-wk:
- Added is_model_int8 to detect int8 models by filename or layer property.
- Refactored benchmarking to use run_benchmark_if_supported, which loads model files, checks for int8, and skips GPU benchmarking for int8 models with a clear message.
- Applies to both default and user-specified models(param).

### DIFF
--- a/benchmark/benchncnn.cpp
+++ b/benchmark/benchncnn.cpp
@@ -48,6 +48,14 @@ static ncnn::VkAllocator* g_staging_vkallocator = 0;
 
 void benchmark(const char* comment, const std::vector<ncnn::Mat>& _in, const ncnn::Option& opt, bool fixed_path = true)
 {
+    // Skip if int8 model name and using GPU
+    if (opt.use_vulkan_compute && strstr(comment, "int8") != NULL)
+    {
+        if (!fixed_path)
+            fprintf(stderr, "%20s  skipped (int8+GPU not supported)\n", comment);
+        return;
+    }
+
     g_blob_pool_allocator.clear();
     g_workspace_pool_allocator.clear();
 


tencent-adm:
[CLA assistant check](https://cla.opensource.tencent.com/Tencent/ncnn?pullRequest=6174) <br/>All committers have signed the CLA.

benchmark/benchncnn.cpp
```diff
@@ -380,82 +419,52 @@ int main(int argc, char** argv)
...
+        run_benchmark_if_supported("mobilenet_ssd", ncnn::Mat(300, 300, 3), opt);
+        run_benchmark_if_supported("mobilenet_ssd_int8", ncnn::Mat(300, 300, 3), opt);
+        run_benchmark_if_supported("mobilenet_yolo", ncnn::Mat(416, 416, 3), opt);
+        run_benchmark_if_supported("mobilenetv2_yolov3", ncnn::Mat(352, 352, 3), opt);
+        run_benchmark_if_supported("yolov4-tiny", ncnn::Mat(416, 416, 3), opt);
+        run_benchmark_if_supported("nanodet_m", ncnn::Mat(320, 320, 3), opt);
+        run_benchmark_if_supported("yolo-fastest-1.1", ncnn::Mat(320, 320, 3), opt);
+        run_benchmark_if_supported("yolo-fastestv2", ncnn::Mat(352, 352, 3), opt);
+        run_benchmark_if_supported("vision_transformer", ncnn::Mat(384, 384, 3), opt);
+        run_benchmark_if_supported("FastestDet", ncnn::Mat(352, 352, 3), opt);
```
nihui: Add the logic of `run_benchmark_if_supported` directly to the benchmark() function so that we don’t need to modify the call here.run_benchmark_if_supported

nihui (changes requested):

benchmark/benchncnn.cpp
```diff
@@ -241,6 +241,45 @@ static std::vector<ncnn::Mat> parse_shape_list(char* s)
...
+    // Check filename convention
+    if (strstr(model_name, "int8") != nullptr)
+        return true;
+
+    // Check actual model layers
+    for (size_t i = 0; i < net.layers().size(); i++)
+    {
+        const ncnn::Layer* layer = net.layers()[i];
+        if (layer->support_int8_storage)
+            return true;
```
nihui: `support_int8_storage` indicates that this layer supports int8 type input, but it does not mean that the current layer must not use int8 calculation, considering dynamic quantization, such as rnn. However, I think the logic here is OK, as a supplementary judgment

benchmark/benchncnn.cpp
```diff
@@ -90,6 +90,13 @@ void benchmark(const char* comment, const std::vector<ncnn::Mat>& _in, const ncn
     DataReaderFromEmpty dr;
     net.load_model(dr);
 
+    // Check if model is int8 and skip on GPU
+    if (opt.use_vulkan_compute && is_model_int8(net, comment))
+    {
+        fprintf(stderr, "%20s  skipped (int8+GPU not supported)\n", comment);
+        return;
+    }
+
```
nihui: Since only the name pattern is checked, move these lines to the beginning of `benchmark()` for skipping early without loading model, and inline `is_model_int8()` call which is simple enough.

In `fixed_path` mode, skip silently.

nihui (changes requested):

benchmark/benchncnn.cpp
```diff
@@ -48,6 +48,14 @@ static ncnn::VkAllocator* g_staging_vkallocator = 0;
 
 void benchmark(const char* comment, const std::vector<ncnn::Mat>& _in, const ncnn::Option& opt, bool fixed_path = true)
 {
+    // Skip if int8 model name and using GPU
+    if (opt.use_vulkan_compute && strstr(comment, "int8") != nullptr)
```
nihui: use `NULL` for cpp03 compatible

nihui (changes requested):

nihui (approved):

nihui:

PR #6179, opened 2025-07-12
vulkan unaryop unified elempack shader

weikangqi:
- [x] impl vulkan unaryop unified elempack shader


tencent-adm:
[CLA assistant check](https://cla.opensource.tencent.com/Tencent/ncnn?pullRequest=6179) <br/>All committers have signed the CLA.

nihui:
aha, to trigger code-format, push another dummy commit (such as inserting an empty line)

src/layer/vulkan/shader/unaryop.comp
```diff
@@ -3,40 +3,27 @@
...
 
 layout (push_constant) uniform parameter
 {
-    int dims;
-    int w;
-    int h;
-    int c;
-    int cstep;
+    uint n;
+    int op_type;
```
nihui: `op_type` is always known at compile-time, it is no necessary to be a push constant

src/layer/vulkan/shader/unaryop.comp
```diff
@@ -3,40 +3,27 @@
 
 #version 450
 
-layout (constant_id = 0) const int op_type = 0;
+layout (constant_id = 0) const uint n = 0;
+layout (constant_id = 1) const int op_type = 0;
```
nihui: change the order so specialization constant without a push constant fallback has lower id than others

nihui (changes requested):

nihui:

PR #6183, opened 2025-07-13
wip: [​fix(cpu): add support for CPUs with more than 64 cores]

hungryDodo:
Addresses two critical issues with ncnn on systems with more than 64 CPU cores:

1. Refactoring ncnn::CpuSet to support more cores:

   - On Linux platforms, the original cpu_set_t implementation had a fixed-size limitation. It has been refactored to use CPU_ALLOC for dynamic memory allocation, along with macros with the _S suffix (e.g., CPU_SET_S) for operations. This ensures proper support for affinity settings on systems with more than CPU_SETSIZE (typically 1024) cores.

2. Fixing race conditions in global CPU info initialization:

   - A race condition was discovered in the try_initialize_global_cpu_info function under multithreaded environments (e.g., when benchncnn runs on 128 cores), which could lead to multiple invocations of the initialization function and cause program crashes.

   - By replacing the original non-thread-safe flag check with pthread_once, the initialization of global CPU information is now thread-safe and guaranteed to execute only once during the program's lifecycle.

### DIFF
--- a/src/cpu.cpp
+++ b/src/cpu.cpp
@@ -1452,25 +1452,25 @@ static int set_sched_affinity(const ncnn::CpuSet& thread_affinity_mask)
             PROCESSOR_NUMBER proc_num;
             proc_num.Group = group_index;
             proc_num.Number = proc_index_in_group;
-            
+
             // We need a way to map from PROCESSOR_NUMBER to a global linear CPU index.
             // This is complex without a pre-built map.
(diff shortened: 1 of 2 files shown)


tencent-adm:
[CLA assistant check](https://cla.opensource.tencent.com/Tencent/ncnn?pullRequest=6183) <br/>All committers have signed the CLA.

PR #6184, opened 2025-07-13
fix: adjust platform-specific logic in cpu.cpp

hungryDodo:


nihui:
On the same machine, `try_initialize_global_cpu_info` always produces consistent results. Even if there may be data race in threaded calls, the results are always consistent.

pthread once is not needed

nihui:
Duplicate of #5681

hungryDodo:
Thanks for pointing that out. 🐱

PR #6197, opened 2025-07-21
support cpu count > 64 issue6142

DaChengTechnology:
add MutiThread Module


tencent-adm:
[CLA assistant check](https://cla.opensource.tencent.com/Tencent/ncnn?pullRequest=6197) <br/>All committers have signed the CLA.

DaChengTechnology:
some layer not support.It is necessary for @nihui  to determine whether he can accept this style of code.

DaChengTechnology:
> The goal of issue 6142 is to add CPU topology information retrieval and thread affinity handling. You should not modify the way OpenMP is used in the layer.

Windows  thread affinity handling is contian two pamaes, the one indicate that cpu groups,other one indicate that cpu core mark.
how do you use  that on OpenMP`s program?

DaChengTechnology:
now, ncnn lib use openMP maybe not in thread affinity.

PR #6254, opened 2025-08-14
esp32 build doc: advise not running install.sh with sudo

Willaaaaaaa:
The existing doc recommended users to run `sudo ./install.sh`, which can lead to `esp-idf` being install in `/root/.espressif` rather than `/home/<username>/.espressif`.


nihui (approved):

nihui:

PR #6270, opened 2025-08-21
drop vulkan shader pack8

nihui:
The pack8 shader code is designed to round up the fp16 type to 128 bits, thus achieving the correct performance boost for image storage on the old Adreno graphics

As image storage is deprecated, simply pack8 won't provide any performance gains

Let's wait for the vulkan long vector extension

- [x] drop sfpvec8/afpvec8/sintvec8/aintvec8 glsl extension


tencent-adm:
[CLA assistant check](https://cla.opensource.tencent.com/Tencent/ncnn?pullRequest=6270) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.opensource.tencent.com/Tencent/ncnn?pullRequest=6270) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.opensource.tencent.com/check/Tencent/ncnn?pullRequest=6270) it.</sub>

PR #6306, opened 2025-09-09
shader: use #extension GL_GOOGLE_include_directive: enable for compatibility

whyb:
## shader changes

This PR modifies the directive in ```/src/layer/vulkan/shader/***.comp``` from ```#extension GL_GOOGLE_include_directive : require``` to ```#extension GL_GOOGLE_include_directive : enable```  for compatibility. 

## why do this change

Some Vulkan drivers—especially those not based on Google's implementation—do not support the require level for this extension, which can lead to shader compilation failures. Switching to enable improves compatibility across platforms while preserving functionality where the extension is supported.
This change does not affect shader logic and only adjusts the extension declaration to be more tolerant.

### DIFF
--- a/src/layer/vulkan/shader/convolution_1x1s1d1_cm.comp
+++ b/src/layer/vulkan/shader/convolution_1x1s1d1_cm.comp
@@ -3,7 +3,7 @@
 
 #version 450
 
-#extension GL_GOOGLE_include_directive: require
+#extension GL_GOOGLE_include_directive: enable
 #include "vulkan_activation.comp"
 
 #extension GL_EXT_control_flow_attributes: require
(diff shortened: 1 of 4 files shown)


tencent-adm:
[CLA assistant check](https://cla.opensource.tencent.com/Tencent/ncnn?pullRequest=6306) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.opensource.tencent.com/Tencent/ncnn?pullRequest=6306) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.opensource.tencent.com/check/Tencent/ncnn?pullRequest=6306) it.</sub>

nihui:
`GL_GOOGLE_include_directive` is always supported by glslang.

PR #6315, opened 2025-09-11
fix format specifier for size_t to support 64-bit values

whyb:
## why do this change
This PR fixes incorrect format specifiers in a diagnostic fprintf statement that prints tensor and model input counts. Previously, ```%ld``` was used to print values returned by ```std::vector::size()```, which may lead to incorrect output on platforms where ```size_t``` is 64-bit. The format specifier has been updated to ```%lld``` to ensure correct printing of 64-bit values across platforms.
so it pr could improves cross-platform correctness of runtime logs.
Prevents misleading output when input count exceeds 2³¹ on 64-bit builds.

## changes
replaced ```%ld``` with ```%lld``` in ```fprintf(stderr, ...)``` to match size_t width on 64-bit systems.

## standards reference
This change aligns with the format specifier definitions introduced in the ISO/IEC 9899:1999 (C99) standard, which formally added support for the ```long long int``` type and its corresponding printf format specifier ```%lld```. Using ```%lld``` ensures correct representation of 64-bit integer values across platforms that conform to C99 or later, improving portability and diagnostic accuracy.
In stl, the definition of size_t is always unsigned __int64, ```unsigned __int64 == unsigned long long int```
see page 276 of the C99 standard: 
[C99 standard document.pdf](https://github.com/user-attachments/files/22269100/C99_standard.pdf)

图片

### DIFF
--- a/benchmark/benchncnn.cpp
+++ b/benchmark/benchncnn.cpp
@@ -242,7 +242,7 @@ static std::vector<ncnn::Mat> parse_shape_list(char* s)
             mats.push_back(ncnn::Mat(shape[0]));
             break;
         default:
-            fprintf(stderr, "unsupported input shape size %lld\n", shape.size());
+            fprintf(stderr, "unsupported input shape size %zu\n", shape.size());
             break;
         }
     }


tencent-adm:
[CLA assistant check](https://cla.opensource.tencent.com/Tencent/ncnn?pullRequest=6315) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.opensource.tencent.com/Tencent/ncnn?pullRequest=6315) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.opensource.tencent.com/check/Tencent/ncnn?pullRequest=6315) it.</sub>

whyb:
> use `%zu` or `%d` with `(int)xxx.size()`

I was apply copilot's good suggestion now.👍

nihui:

PR #6363, opened 2025-10-21
Update build instructions to include 'make install'

roachsinai:
Added 'make install' command to build instructions for AArch32 and AArch64 targets.


nihui:
fine

nihui:

PR #6365, opened 2025-10-22
feat: rms vulkan shader

futz12:
prepare for llm

change:

add rmsnorm shader
unifi naive cpu code style between rmsnorm and layernorm
add notice in layernorm shader because rmsnorm reuse some shader of layernorm


nihui:

PR #6368, opened 2025-10-22
fix: softmax x86 miss tail data feat: use rcp to replace 1/x

futz12:
fix: softmax x86 miss tail
feat: use rcp to replace 1/x


nihui (approved):

nihui:

PR #6378, opened 2025-10-30
Ensure platform.h is included before other header files

quink-black:
When _WIN32_WINNT isn't set in compiler flags, it can be defined by platform.h or windows.h, depends on which one is included first. Then we get different value of _WIN32_WINNT.

The conditional compile of ncnn::Mutex and so on depends on the value of _WIN32_WINNT. Different define of _WIN32_WINNT leading to two implementations of ncnn::Mutex, which is ODR violation. ncnn can crash at runtime.

Add a check in platform.h to ensure _WINDOWS_ macro isn't defined. If windows.h is included before platform.h, trigger a compilation failure to enforce that platform.h is included first.

Check all places that include platform.h and adjust there including order, or remove redundant includes for platform.h.

### DIFF
--- a/src/mat_pixel_affine.cpp
+++ b/src/mat_pixel_affine.cpp
@@ -7,7 +7,6 @@
 #endif // __ARM_NEON
 #include <limits.h>
 
-
 namespace ncnn {
 
 #if NCNN_PIXEL_AFFINE


tencent-adm:
[CLA assistant check](https://cla.opensource.tencent.com/Tencent/ncnn?pullRequest=6378) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.opensource.tencent.com/Tencent/ncnn?pullRequest=6378) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.opensource.tencent.com/check/Tencent/ncnn?pullRequest=6378) it.</sub>

PR #6284, opened 2025-08-26
[WIP] flexible coopmat optimization for vulkan gemm

nihui:
- [ ] unroll subgroup k
- [ ] fp16 packed support
- [ ] gemm benchmark

---

This pull request adds support for using Vulkan cooperative matrix operations in the `Gemm_vulkan` layer, optimizing matrix multiplication when the hardware and options allow. The main changes introduce logic to detect support for cooperative matrices, configure optimal tiling and unrolling parameters, and dispatch the computation accordingly.

**Cooperative matrix support and pipeline configuration:**

* Added new member variables to `Gemm_vulkan` for cooperative matrix support, including flags and tiling/unrolling parameters (`use_cooperative_matrix`, `coopmat_M`, `coopmat_N`, `coopmat_K`, `UNROLL_SG_M`, `UNROLL_SG_N`, `UNROLL_SG_K`, `UNROLL_WG_M`, `UNROLL_WG_N`) (`src/layer/vulkan/gemm_vulkan.h`).
* Initialized these new member variables in the constructor to ensure safe defaults (`src/layer/vulkan/gemm_vulkan.cpp`).

**Pipeline creation and specialization:**

* In `create_pipeline`, added logic to enable cooperative matrix usage if supported and requested, query optimal tile sizes, compute unrolling factors, and set up pipeline specialization constants for the cooperative matrix shader (`src/layer/vulkan/gemm_vulkan.cpp`).
* Adjusted pipeline creation to select the cooperative matrix shader and configure subgroup/local sizes accordingly when cooperative matrix is used (`src/layer/vulkan/gemm_vulkan.cpp`) [[1]](diffhunk://#diff-2b03b12c002e81276043a93210a046458d1bc4f74bc44273f4f6a15f4219fad4R72-R126) [[2]](diffhunk://#diff-2b03b12c002e81276043a93210a046458d1bc4f74bc44273f4f6a15f4219fad4R163).

**Forward pass dispatch logic:**

* Updated the `forward` method to dispatch the cooperative matrix pipeline with the correct workgroup and tiling configuration, falling back to the standard pipeline otherwise (`src/layer/vulkan/gemm_vulkan.cpp`).

PR #6400, opened 2025-11-12
double buffer for gemm

futz12:


nihui:
there is one interesting flops test in https://github.com/Tencent/ncnn/pull/6284

PR #6436, opened 2025-12-02
riscv: implement rvv optimized tan operator

ihb2032:
### Summary
This PR implements RVV optimization for the `tan` unary operator.

### Verification
- Device: SG2044 (Sophgo)
- Status: Passed `tests/test_unaryop` verification.

PR #6460, opened 2025-12-21
riscv: implement rvv optimized tan operator

ihb2032:
This is a follow-up to #6436. I've applied the macro fix as suggested by @nihui to exclude xtheadvector, ensuring precision while keeping the optimization for standard RVV 1.0.

### DIFF
--- a/src/layer/riscv/unaryop_riscv.cpp
+++ b/src/layer/riscv/unaryop_riscv.cpp
@@ -164,7 +164,7 @@ struct unary_op_tan
 {
     vfloat32m8_t operator()(const vfloat32m8_t& x, const size_t& vl) const
     {
-        // TODO rvv optimize
+#if __riscv_xtheadvector
         std::vector<float> tmp(vl);
         __riscv_vse32_v_f32m8(tmp.data(), x, vl);
         for (size_t i = 0; i < vl; i++)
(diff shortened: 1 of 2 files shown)


src/layer/riscv/unaryop_riscv.cpp
```diff
@@ -164,14 +164,9 @@ struct unary_op_tan
...
-        std::vector<float> tmp(vl);
-        __riscv_vse32_v_f32m8(tmp.data(), x, vl);
-        for (size_t i = 0; i < vl; i++)
-        {
-            tmp[i] = tanf(tmp[i]);
-        }
-        return __riscv_vle32_v_f32m8(tmp.data(), vl);
+        vfloat32m8_t sin_x, cos_x;
+        sincos_ps(x, &sin_x, &cos_x, vl);
+        return __riscv_vfdiv_vv_f32m8(sin_x, cos_x, vl);
```
nihui: exclude __riscv_xtheadvector for test failure on fp32

nihui (changes requested):

nihui (changes requested):
```cpp
vfloat32m8_t operator()(const vfloat32m8_t& x, const size_t& vl) const
{
#if __riscv_xtheadvector
    // fallback
#else
    // rvv opt
#endif
}
```

nihui (approved):

nihui:

PR #6473, opened 2026-01-01
selu vulkan shader

futz12:


nihui:

PR #6483, opened 2026-01-03
softmax onestage vulkan shader

futz12:


futz12:
@copilot How do you think of this pr

PR #6496, opened 2026-01-08
Fix windown header incldue for platform.h.in

L1ghtError:
Because we check for the presence of `_WIN32_WINNT` before including the actual `#include <windows.h>`, we will fall back to `_WIN32_WINNT 0x0501` if the user code does not include `#include <windows.h>` before including ncnn.

I encountered a problem when the main code includes `windows.h` but some library `.dlls` do not. In that case, `PoolAllocatorPrivate` will allocate memory on the stack for `ncnn::Mutex` using `SRWLOCK` (because the main code has included `windows.h`), but the constructor is called with `InitializeCriticalSection` (because in the shared library I did not include `windows.h`, so `ncnn` falls back to `Windows XP` behavior), which results in memory corruption.

Example:
image

image


tencent-adm:
[CLA assistant check](https://cla.opensource.tencent.com/Tencent/ncnn?pullRequest=6496) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.opensource.tencent.com/Tencent/ncnn?pullRequest=6496) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.opensource.tencent.com/check/Tencent/ncnn?pullRequest=6496) it.</sub>

PR #6511, opened 2026-01-14
pnnx tnn2pnnx support Flatten operator in TNN model conversion

Missmiaom:
## Summary
Add missing `Flatten` operator mapping in TNN to PNNX conversion.

## Changes
- Add `tnn.Flatten` to `Flatten` type conversion in operator replacement pass
- Place it under the unary operators section for better organization

## Motivation
Currently, TNN models containing `Flatten` layers fail to convert properly because the operator type mapping is missing. This PR fixes the issue by adding the necessary mapping.

## Testing
- Tested with TNN model containing Flatten operator
- Conversion completes successfully with this change

### DIFF
--- a/tools/pnnx/src/load_tnn.cpp
+++ b/tools/pnnx/src/load_tnn.cpp
@@ -613,6 +613,7 @@ int load_tnn(const std::string& tnnpath, Graph& pnnx_graph)
     {
         // unary
         if (op->type == "tnn.Erf") op->type = "aten::erf";
+        if (op->type == "tnn.Flatten") op->type = "Flatten";
         if (op->type == "tnn.Log") op->type = "aten::log";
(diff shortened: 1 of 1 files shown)


tencent-adm:
[CLA assistant check](https://cla.opensource.tencent.com/Tencent/ncnn?pullRequest=6511) <br/>All committers have signed the CLA.

nihui (changes requested):
`Flatten` is not unaryop

you shall add the conversion in `pnnx/src/pass_level2/torch_flatten.cpp`

you can refer how `tnn.Reshape` handled in `pnnx/src/pass_level2/Tensor_reshape.cpp`

Missmiaom:
Thank you, I found the problem and closed this pr

PR #6513, opened 2026-01-15
pnnx: add TNN Flatten operator support in pass_level2

Missmiaom:
## Summary

Add support for TNN `Flatten` operator in PNNX pass_level2 conversion.

## Problem

Currently, TNN models containing `Flatten` layers fail to convert to PNNX/NCNN format because the operator mapping is missing in pass_level2.

## Solution

Implement `torch_flatten_tnn` pass in `torch_flatten.cpp` following the same pattern as the existing `torch_flatten_onnx` pass.

**Key changes:**
- Add `torch_flatten_tnn` class to match `tnn.Flatten` operator
- Map TNN `arg0` parameter to PyTorch `start_dim` parameter
- Set `end_dim=-1` to flatten to the last dimension
- Use default `start_dim=1` when parameter is not specified


nihui:

PR #6516, opened 2026-01-16
Add a CMake option to use system-provided pybind11

Integral-Tech:
This option enables the use of the pybind11 package available in the official repositories of Linux distributions.


tencent-adm:
[CLA assistant check](https://cla.opensource.tencent.com/Tencent/ncnn?pullRequest=6516) <br/>All committers have signed the CLA.

Integral-Tech:
@nihui IMO, the suggestion from Copilot, which uses the `REQUIRED` keyword in the `find_package()` call, is not optimal. It would be better to fallback to the submodule if the pybind11 package is not found, similar to how the glslang finding works.

I have just pushed a new commit to include the latest changes :)

nihui:

PR #6549, opened 2026-02-15
more opearation support for binaryop

futz12:
1. finish todo of binaryop


src/layer/binaryop.cpp
```diff
@@ -240,6 +308,14 @@ static void binary_op_broadcast(const Mat& a, const Mat& b, Mat& c, int op_type,
...
     if (op_type == BinaryOp::Operation_ATAN2) return binary_op_broadcast<binary_op_atan2>(a, b, c, opt);
     if (op_type == BinaryOp::Operation_RATAN2) return binary_op_broadcast<binary_op_atan2>(b, a, c, opt);
+    if (op_type == BinaryOp::Operation_FMOD) return binary_op_broadcast<binary_op_fmod>(a, b, c, opt);
+    if (op_type == BinaryOp::Operation_RFMOD) return binary_op_broadcast<binary_op_rfmod>(a, b, c, opt);
+    if (op_type == BinaryOp::Operation_LOGADDEXP) return binary_op_broadcast<binary_op_logaddexp>(a, b, c, opt);
+    if (op_type == BinaryOp::Operation_RLOGADDEXP) return binary_op_broadcast<binary_op_rlogaddexp>(a, b, c, opt);
+    if (op_type == BinaryOp::Operation_FLOOR_DIVIDE) return binary_op_broadcast<binary_op_floor_divide>(a, b, c, opt);
+    if (op_type == BinaryOp::Operation_RFLOOR_DIVIDE) return binary_op_broadcast<binary_op_rfloor_divide>(a, b, c, opt);
+    if (op_type == BinaryOp::Operation_REMAINDER) return binary_op_broadcast<binary_op_remainder>(a, b, c, opt);
+    if (op_type == BinaryOp::Operation_RREMAINDER) return binary_op_broadcast<binary_op_rremainder>(a, b, c, opt);
```
nihui: In the generic binaryop.cpp, the existing reverse operations (RSUB, RDIV, RPOW, RATAN2) are dispatched by swapping a and b and using the forward functor:
```
  if (op_type == BinaryOp::Operation_RPOW) return binary_op_broadcast<binary_op_pow>(b, a, c, opt);
  if (op_type == BinaryOp::Operation_RATAN2) return binary_op_broadcast<binary_op_atan2>(b, a, c, opt);
```
  But the new reverse operations use the reverse functor with the original argument order instead of the forward functor with swapped arguments:
```
  if (op_type == BinaryOp::Operation_RFMOD) return binary_op_broadcast<binary_op_rfmod>(a, b, c, opt);
  if (op_type == BinaryOp::Operation_RFLOOR_DIVIDE) return binary_op_broadcast<binary_op_rfloor_divide>(a, b, c, opt);
  if (op_type == BinaryOp::Operation_RREMAINDER) return binary_op_broadcast<binary_op_rremainder>(a, b, c, opt);
  if (op_type == BinaryOp::Operation_RLOGADDEXP) return binary_op_broadcast<binary_op_rlogaddexp>(a, b, c, opt);
```
  While this is functionally equivalent for the scalar element-wise case (rfmod(x,y) == fmod(y,x)), it's an inconsistency with the established pattern. The existing pattern binary_op_broadcast<forward_op>(b, a, ...) is more efficient
  because it avoids instantiating a separate reverse template — the broadcast logic handles the swap. This doubles template instantiations unnecessarily for these new operations.

  Recommendation: Follow the existing pattern:
```
  if (op_type == BinaryOp::Operation_RFMOD) return binary_op_broadcast<binary_op_fmod>(b, a, c, opt);
```

src/layer/binaryop.cpp
```diff
@@ -226,6 +226,74 @@ struct binary_op_ratan2
...
+
+struct binary_op_rlogaddexp
+{
+    float operator()(const float& x, const float& y) const
+    {
+        float max_xy = std::max(y, x);
+        float min_xy = std::min(y, x);
+        return (float)(max_xy + log1pf(expf(min_xy - max_xy)));
+    }
+};
```
nihui, line 253: logaddexp(x, y) = log(exp(x) + exp(y)) is a commutative operation — logaddexp(x, y) == logaddexp(y, x). Having Operation_RLOGADDEXP as a separate enum value is mathematically redundant. The binary_op_rlogaddexp functor computes
  max(y,x) + log1p(exp(min(y,x) - max(y,x))) which is identical to max(x,y) + log1p(exp(min(x,y) - max(x,y))).

  While this doesn't cause incorrect results, it wastes enum space, increases code size across all backends, and is conceptually misleading. If the intent is to keep symmetry with other operations for the get_reverse_op_type function, a
   comment explaining this would be helpful. Or get_reverse_op_type could simply return Operation_LOGADDEXP for both directions.

src/layer/arm/neon_mathfun.h
```diff
@@ -404,6 +404,91 @@ static inline float32x4_t atan2_ps(float32x4_t a, float32x4_t b)
     return vld1q_f32(tmpx);
 }
 
+static inline float32x4_t trunc_ps(const float32x4_t& x)
+{
+    // truncate toward zero
+    int32x4_t xi = vcvtq_s32_f32(x);
+    return vcvtq_f32_s32(xi);
+}
```
nihui, line 416: On aarch64, NEON provides vrndq_f32 for truncation toward zero, but the implementation uses the less precise int-conversion approach:
```  static inline float32x4_t trunc_ps(const float32x4_t& x) {
      int32x4_t xi = vcvtq_s32_f32(x);
      return vcvtq_f32_s32(xi);
  }
```
  Compare with floor_ps which correctly uses the hardware instruction on aarch64:
```  static inline float32x4_t floor_ps(const float32x4_t& x) {
  #if __aarch64__
      return vrndmq_f32(x);
  #else
      // fallback...
  #endif
  }
```

nihui:

DickyQi:
Operation_LOGADDEXP vulkan shader failed compile on fp16 mode
I think it should be change link this:
if (op_type == 14) res = max(v1, v2) + log(1.0 + exp(min(v1, v2) - max(v1, v2)));  -->
if (op_type == 14) res = max(v1, v2) + log(afp(1.0) + exp(min(v1, v2) - max(v1, v2)));

nihui:
> Operation_LOGADDEXP vchange  shader failed compile on fp16 mode I think it should be change link this: if (op_type == 14) res = max(v1, v2) + log(1.0 + exp(min(v1, v2) - max(v1, v2))); --> if (op_type == 14) res = max(v1, v2) + log(afp(1.0) + exp(min(v1, v2) - max(v1, v2)));

feel free to drop a pull request change :D

nihui:
> Operation_LOGADDEXP vulkan shader failed compile on fp16 mode I think it should be change link this: if (op_type == 14) res = max(v1, v2) + log(1.0 + exp(min(v1, v2) - max(v1, v2))); --> if (op_type == 14) res = max(v1, v2) + log(afp(1.0) + exp(min(v1, v2) - max(v1, v2)));

https://github.com/Tencent/ncnn/pull/6602

PR #6550, opened 2026-02-19
Add CompareOp

futz12:
support for torch.lt/gt/le/ge/eq/ne


futz12:
https://github.com/Tencent/ncnn/pull/6551

PR #6584, opened 2026-03-08
x86: add AbsVal_x86 with bf16 storage support

nihui:
Use floating-point AND (SSE2/AVX/AVX-512F) to clear the sign bit of each bf16/fp32 element. For bf16, reinterpret pairs as float and mask with 0x7fff7fff to clear both sign bits in one 32-bit AND -- no AVX2 integer instructions needed, no fp32 round-trip required.

### DIFF
--- a/src/layer/x86/absval_x86.cpp
+++ b/src/layer/x86/absval_x86.cpp
@@ -1,0 +1,157 @@
+// Copyright 2026 Tencent
+// SPDX-License-Identifier: BSD-3-Clause
+
+#include "absval_x86.h"
+
+#if __SSE2__
+#include <emmintrin.h>
+#if __AVX__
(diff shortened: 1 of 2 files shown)


tencent-adm:
[CLA assistant check](https://cla.opensource.tencent.com/Tencent/ncnn?pullRequest=6584) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.opensource.tencent.com/Tencent/ncnn?pullRequest=6584) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.opensource.tencent.com/check/Tencent/ncnn?pullRequest=6584) it.</sub>

PR #6636, opened 2026-04-04
RVV 1.0 Quantize Layer

Deepdive543443:
I was looking into how to make int8 optimization happen on RV, then realize we need packed & fp16s/a quantization, so here we go, Quantize layer for RVV1.0!

WIP as I'm not sure we're all happy with the current approach. I'm also not sure if it could works alongside with THead Vector. 

- [x] Packed int8 quantize
- [x] Packed fp16s/a int8 quantize
- [x] More unit test to cover longer VLEN

### DIFF
--- a/src/layer/riscv/quantize_riscv.cpp
+++ b/src/layer/riscv/quantize_riscv.cpp
@@ -33,7 +33,6 @@ static void quantize(const float* ptr, signed char* s8ptr, const Mat& scale_data
     const int size = elemcount * elempack;
     float scale = scale_data[0];
 
-    int i = 0;
 #if __riscv_vector
     int n = size;
     while (n > 0)
@@ -47,15 +46,14 @@ static void quantize(const float* ptr, signed char* s8ptr, const Mat& scale_data
         s8ptr += vl;
         n -= vl;
     }
-
-    i += (size - n);
-#endif
-    for (; i < size; i++)
+#else  // __riscv_zvfh
+    for (int i = 0; i < size; i++)
     {
         *s8ptr = float2int8(*ptr * scale);
         ptr++;
         s8ptr++;
     }
+#endif // __riscv_zvfh
 }
 
 #if __riscv_vector
@@ -296,6 +294,7 @@ int Quantize_riscv::forward(const Mat& bottom_blob, Mat& top_blob, const Option&
         }
         if (elempack == packn && out_elempack == 1)
         {
+            #pragma omp parallel for num_threads(opt.num_threads)
             for (int q = 0; q < channels; q++)
             {
                 const float* ptr = bottom_blob.channel(q);
--- a/src/layer/riscv/quantize_riscv.cpp
+++ b/src/layer/riscv/quantize_riscv.cpp
@@ -246,7 +246,7 @@ int Quantize_riscv::forward(const Mat& bottom_blob, Mat& top_blob, const Option&
             }
         }
 #endif // __riscv_vector
-        if (elempack == out_elempack)
+        if (elempack == 1 && out_elempack == 1)
         {
             #pragma omp parallel for num_threads(opt.num_threads)
             for (int i = 0; i < h; i++)
@@ -307,7 +307,7 @@ int Quantize_riscv::forward(const Mat& bottom_blob, Mat& top_blob, const Option&
             }
         }
 #endif // __riscv_vector
-        if (elempack == out_elempack)
+        if (elempack == 1 && out_elempack == 1)
         {
             #pragma omp parallel for num_threads(opt.num_threads)
             for (int q = 0; q < channels; q++)
--- a/src/layer/riscv/quantize_riscv.cpp
+++ b/src/layer/riscv/quantize_riscv.cpp
@@ -35,24 +35,12 @@ static void quantize(const float* ptr, signed char* s8ptr, const Mat& scale_data
 
     int i = 0;
 #if __riscv_vector
-    const size_t vlm1 = __riscv_vsetvlmax_e32m1();
-    vfloat32m8_t _scale;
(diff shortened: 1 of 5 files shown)


nihui:
```
  Issue 2: Mismatched conditional guards for forward_fp16sa (real bug)                                                                                
                                                                                                                                                      
  Header (quantize_riscv.h:373):
  #if NCNN_ZFH                                                                                                                                        
      int forward_fp16sa(...);                                    
  #endif                                                                                                                                              
                                                                  
  Implementation (quantize_riscv_zfh.cpp:844):
  #if __riscv_zvfh    // <-- different guard!
  int Quantize_riscv::forward_fp16sa(...)    
```

```
  Issue 4: Dead code in quantize_fp16s per-element scale path
                                                             
  In quantize_riscv_zfh.cpp:416:
  else if (elempack == vlm1)                                                                                                                          
  {                         
      // per-element scale via vlm1-wide load                                                                                                         
  }                                                               
   
  This branch handles per-element scale broadcasting, but elempack is always 1 in all callers of quantize_fp16s (it's called with elempack=1 from
  forward_fp16s regardless of dims). The only way this activates is if vlm1 == 1, which would mean VLEN=32 bits — violating the RISC-V Vector spec    
  minimum of VLEN≥128. So this is dead code for any spec-compliant target. The branch should probably be else if (scale_data.w == elempack) or removed
   entirely.                                                                                                                                          
```

```
  Positive observations
                                                                                                                                                      
  - The overall structure correctly mirrors the ARM implementation pattern.
  - Scalar fallback paths are properly maintained for non-vector targets.                                                                             
  - Threading with #pragma omp parallel for and proper loop partitioning is correct.                                                                  
  - forward_fp16sa (ZVFH) correctly stays in fp16 arithmetic throughout, which is appropriate.
  - forward_fp16s correctly widens to float32 before multiplication, consistent with ARM's approach.                                                  
  - Scale data indexing (.range()) is consistent across dims=2 and dims=3 paths.                                                                      
```

nihui (changes requested):
rename `pack4n` and `pack2n` to `packn_s8`

nihui:
```
  Bug 5 (Medium): quantize() scale vector not initialized for all cases

  After patch 5/6, the quantize() function in quantize_riscv.cpp was simplified. But looking at the final state of quantize():

  static void quantize(const float* ptr, signed char* s8ptr, const Mat& scale_data, int elemcount, int elempack)
  {
      const int size = elemcount * elempack;
      float scale = scale_data[0];
      int i = 0;
  #if __riscv_vector
      int n = size;
      while (n > 0)
      {
          ...
          _v0 = __riscv_vfmul_vf_f32m8(_v0, scale, vl);
          ...
      }
      i += (size - n);
  #endif
      for (; i < size; i++)
      {
          *s8ptr = float2int8(*ptr * scale);
          ...
      }
  }

  This function now always uses scale_data[0] — a single scalar scale. But it's called from the elempack == out_elempack path where scale_data might contain per-channel scales (when scale_data_size > 
  1). The caller extracts the correct scale slice via scale_data.range(...), but quantize() still only uses scale_data[0], which is correct for elempack == 1 but incorrect for packed elements where
  each lane may need a different scale.

  Wait — after patch 5/6, the packed paths (packnto4n, packnto1) still handle multi-scale correctly. The quantize() function is only called for elempack == out_elempack, and since the simplified
  version uses scale_data[0] as a scalar, this is only correct when elempack == 1 (and the caller passes a 1-element scale slice) or when scale_data_size == 1 (single global scale). When elempack > 1
  and per-channel scales are used, the caller passes scale_data.range(i * elempack, elempack), but quantize() only reads [0] — so only the first scale in the pack is used.

  This is a real logic bug: for the elempack == out_elempack path when elempack > 1 and scale_data_size > 1, all elements in the pack are quantized with the same scale (the first one) instead of their
   respective per-channel scales.

  Recommendation: Restore the per-element scale handling in quantize() for the packed case, or ensure this function is only called with elempack == 1.
```

nihui:
```
  Bug 6 (Low): Dead i += (size - n) after vectorized loop

  File: quantize_riscv.cpp, quantize():

  int n = size;
  while (n > 0) { ... n -= vl; }
  i += (size - n);  // n is always 0 here

  After the while loop, n == 0 (since the loop terminates when n <= 0), so i += (size - n) simplifies to i += size. This makes the scalar fallback loop for (; i < size; ...) never execute. The scalar fallback is dead code when
  __riscv_vector is enabled.

  This is technically correct (the vector loop handles all elements), but the pattern is misleading. Same issue in quantize_fp16s() and quantize_fp16sa().

  ---
  Bug 7 (Low): Missing #pragma omp parallel for in dims==3 packnto1 path

  File: quantize_riscv.cpp:

  if (elempack == packn && out_elempack == 1)
  {
      for (int q = 0; q < channels; q++)  // <-- no #pragma omp
      {

  The dims==3, elempack==packn, out_elempack==1 path lacks OpenMP parallelization, while all other equivalent paths in dims==2 and dims==3 have it. The same missing pragma exists in quantize_riscv_zfh.cpp for both forward_fp16s and
  forward_fp16sa in the dims==3 packnto1 paths.

  However, this might be intentional since quantize_packnto1 writes to interleaved output positions (s8ptr[i * stride]), so parallelizing the outer loop would require stride = cstep to be thread-safe. Since different q values write to
  different output channels (top_blob.channel(q * packn)), parallelization should actually be safe. Looks like an oversight.

  ---
  Bug 8 (Low): Missing newline at end of file

  Both quantize_riscv.cpp and quantize_riscv_zfh.cpp are missing a trailing newline after } // namespace ncnn, which will generate compiler warnings with -Wnewline-eof.
```

nihui:

PR #6637, opened 2026-04-04
Add RVV implementation for exp

ihb2032:
Add exp operator implementation using RVV intrinsics.


nihui (changes requested):
Legacy layers from the Caffe era, whether in PyTorch or ONNX, will no longer convert to ncnn's Exp and Log, but instead use UnaryOp. However, since you've already optimized them, I believe necessary unit tests are essential. You need to add code like `test_exp` and `test_log` to your tests; you can refer to other test implementations for examples.

src/layer/riscv/exp_riscv.cpp
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    #pragma omp parallel for num_threads(opt.num_threads)
+    for (int q = 0; q < channels; q++)
+    {
+        float* ptr = bottom_top_blob.channel(q);
+
+        for (int i = 0; i < size; i++)
+        {
+            ptr[i] = powf(base, shift + ptr[i] * scale);
```
nihui, line 104: There is `pow_ps` utility.
It also applies to zvfh.

nihui (changes requested):

nihui (approved):
LGTM

nihui:

PR #6661, opened 2026-04-09
fix fuse_padding_convolution asymmetric padding params being skipped

MollySophia:
```
class PadConvModel(nn.Module):
    def __init__(self):
        super().__init__()
        self.conv = nn.Conv2d(
            in_channels=3, out_channels=4,
            kernel_size=5, stride=2, padding=0, bias=True,
        )

    def forward(self, x):
        # asymmetric padding: left=1, right=2, top=1, bottom=2
        x = F.pad(x, pad=(1, 2, 1, 2), mode='constant', value=0)
        x = self.conv(x)
        x = F.relu(x)
        return x
```

#### Before:
Clipboard_Screenshot_1775714112

#### After:
Clipboard_Screenshot_1775714095

#### Info:
When F.pad(pad=(1,2,1,2)) (left=1, right=2, top=1, bottom=2) is fused into a Convolution layer, pad_top (param 14) can be incorrectly set to 0.

Root cause: the write() function first copies all original Convolution params (including 14=0 from padding=0), then conditionally writes the new fused padding values. The condition new_conv_pad_top != new_conv_pad_left skips writing param 14 when they are equal (e.g. both are 1), leaving the stale 14=0 from the copy.

### DIFF
--- a/tools/pnnx/src/pass_ncnn/fuse_padding_convolution.cpp
+++ b/tools/pnnx/src/pass_ncnn/fuse_padding_convolution.cpp
@@ -115,14 +115,10 @@ pnnx.Output             output      1 0 out
         const int new_conv_pad_right = conv_pad_right + pad_right;
         const int new_conv_pad_bottom = conv_pad_bottom + pad_bottom;
 
-        if (new_conv_pad_left != conv_pad_left)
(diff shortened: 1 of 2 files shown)


nihui (approved):

nihui:

PR #6666, opened 2026-04-10
Add RVV implementation for power

ihb2032:
Add power operator implementation using RVV intrinsics.


nihui (approved):

nihui:

PR #6672, opened 2026-04-12
riscv: add absval fp16sa and simplify rvv abs

ihb2032:


src/layer/riscv/absval_riscv.h
```diff
@@ -18,6 +19,7 @@ class AbsVal_riscv : public AbsVal
 protected:
 #if NCNN_ZFH
     int forward_inplace_fp16s(Mat& bottom_top_blob, const Option& opt) const;
+    int forward_inplace_fp16sa(Mat& bottom_top_blob, const Option& opt) const;
```
nihui, line 22: I don't think there is any difference between fp16s and fp16sa. fp16s path is enough.

nihui (changes requested):

PR #6676, opened 2026-04-13
Add RVV implementation for threshold

ihb2032:
Add threshold operator implementation using RVV intrinsics.


nihui (changes requested):
remove line break changes in tests/CMakeLists.txt

nihui (changes requested):
The changes and new unit tests here support the d-dimensional dimension. You need to add d-dimensional support to src/layer/threshold.cpp as well.

nihui (changes requested):
I don't think it makes any sense to dispatch the fp16s path.
drop forward_fp16s and rename forward_fp16sa to forward_fp16s

nihui (approved):

nihui:

PR #6677, opened 2026-04-13
[bugfix] pnnx: emit Erf layer for expression erf

crafcat7:
# Summary
  Expand fused `erf` expressions into the dedicated `Erf` layer instead of routing
  them through `UnaryOp`, and add torchscript / ONNX coverage to prevent regressions.
  This avoids the unsupported `UnaryOp erf` path while keeping the rest of the
  expression expansion unchanged.


nihui:

PR #6681, opened 2026-04-15
cmake: align riscv xtheadvector probe defaults with src logic

ihb2032:
Fix the top-level RISC-V feature probe so its default option selection matches the priority used in `src/CMakeLists.txt`.


nihui:
compiler can support both v and xthreadvector and we can enable both in single binary

nihui (changes requested):
do not remove existing test() and test2(). append your new test3().

ihb2032:
GitHub does not allow reopening this PR after the branch was recreated, so I have opened #7007 with the same branch.

Addressed the review feedback there: the existing test() and test2() are kept unchanged and the new test3() assembling a real th.vmv1r.v instruction is appended. The option defaults are untouched since v and xtheadvector can coexist in a single binary.

PR #6690, opened 2026-04-21
[OPT] x86: optimize PixelShuffle with SIMD block transpose

crafcat7:
# Summary

Adds an x86-specific implementation of `PixelShuffle` (`src/layer/x86/pixelshuffle_x86.{h,cpp}`) that replaces the original scalar stride-store loop with an SSE/AVX block-transpose, yielding a **~3× speedup** on pure-PixelShuffle workloads for both single- and multi-thread paths, with no regressions on any other layer or model.

# Motivation

The generic `PixelShuffle::forward` in `src/layer/pixelshuffle.cpp` uses a stride store
inside the inner-most loop:

```cpp
for (int j = 0; j < w; j++)
{
    outptr[0] = sptr[0];
    sptr++;
    outptr += upscale_factor; // stride store
}
```

Each write jumps `upscale_factor` floats, which defeats auto-vectorization and causes
repeated dirty cache-line traffic. For common super-resolution networks (ESPCN,
RealSR, waifu2x, etc.) the upscale heads are `r=2` or `r=4`, so an SSE/AVX block
transpose followed by contiguous stores is a natural fit.

# Algorithm

For each `(output_channel p, sub-row sh)` we gather `r` source channel pointers and
walk them in lockstep. Each tile of `vec_width × r` is transposed and written
contiguously:

- `r == 2`: AVX `unpacklo_ps / unpackhi_ps + permute2f128` → 8 columns per step;
  SSE `unpacklo_ps / unpackhi_ps` → 4 columns per step.
- `r == 4`: SSE `_MM_TRANSPOSE4_PS` → 4 columns per step (writes 16 contiguous floats).
- `r == 3` or `r > 4`: fall back to the base scalar implementation (correct but slow;
  rarely used in practice).
- Both `mode=0` (CRD) and `mode=1` (DCR) are supported; only the source channel
  index formula differs.

The new layer keeps `support_packing = false` (inherited from the base), so ncnn's
packing machinery automatically unpacks to `pack1` before `forward` runs; the fast
path stays simple and only handles `pack1 + fp32`.

# Correctness

```
ctest -R test_pixelshuffle --output-on-failure
Passed
```
Covers the existing test cases (`r=1/2/3/4`, `mode=0/1`, assorted channel counts).

# Performance

Environment: Linux / WSL2, AMD Ryzen 7 9800X3D (Zen5, full AVX-512 family), g++,
`-O3 -DNDEBUG`, AVX/AVX2/FMA/AVX-512 all enabled. Measurements from commit
`900ca77` using `benchncnn` with `loop=100` and `taskset -c 0-7`; reported as
the `min` metric (most stable at sub-millisecond workloads).

## Pure-PixelShuffle workloads

| Network | Shape | Threads | Baseline (ms) | Optimized (ms) | Speedup |
|---------|-------|--------:|--------------:|---------------:|--------:|
| `pixelshuffle_demo`         | 256×256×64 → 1024×1024×4 (r=4) | 1 | 0.85 | 0.29 | **2.93×** |
| `pixelshuffle_demo`         | same                           | 8 | 0.28 | 0.09 | **3.11×** |
| `pixelshuffle_stacked_demo` | two stacked r=4                | 1 | 0.97 | 0.32 | **3.03×** |
| `pixelshuffle_stacked_demo` | same                           | 8 | 0.34 | 0.10 | **3.40×** |

## End-to-end super-resolution network (ESPCN topology)

Synthetic ESPCN `benchmark/espcn_demo.param`:
Input 256×256×3 → Conv5×5 (64) → ReLU → Conv3×3 (32) → ReLU → Conv3×3 (48) → PixelShuffle r=4 → 1024×1024×3.

| Threads | Baseline (ms) | Optimized (ms) | Speedup |
|--------:|--------------:|---------------:|--------:|
| 1       | 18.66         | 18.77          | ≈1× (within noise) |
| 8       | 10.96         | 10.12          | 1.08×   |

Conv work dominates this topology (≈4.8 GMACs vs. a pure-data-movement
PixelShuffle). The layer-level speedup is real but its absolute magnitude
(~0.7 ms) is small compared to the conv total. The optimization matters most
in networks where PixelShuffle is a larger fraction of the runtime
(upsampling heads, lightweight SR / style transfer, embedded / single-thread
deployments).

## No regressions

All 35+ built-in benchmark models (squeezenet, mobilenet v1/v2/v3,
shufflenet v1/v2, googlenet, resnet18/50, vgg16, yolo v3/v4-tiny, nanodet,
efficientnet b0 / v2-b0, vision_transformer, etc.) were measured before/after
at both `threads=1` and `threads=8`. All timings are within noise (±3%). None
of these networks use PixelShuffle.

# Files

- `src/layer/x86/pixelshuffle_x86.h` — new header
- `src/layer/x86/pixelshuffle_x86.cpp` — new implementation

### DIFF
--- a/src/layer/x86/pixelshuffle_x86.cpp
+++ b/src/layer/x86/pixelshuffle_x86.cpp
@@ -20,10 +20,6 @@ PixelShuffle_x86::PixelShuffle_x86()
 
 int PixelShuffle_x86::forward(const Mat& bottom_blob, Mat& top_blob, const Option& opt) const
 {
-    int elembits = bottom_blob.elembits();
-    if (elembits != 32)
-        return PixelShuffle::forward(bottom_blob, top_blob, opt);
-
     const int w = bottom_blob.w;
     const int h = bottom_blob.h;
     const int channels = bottom_blob.c;
@@ -33,9 +29,6 @@ int PixelShuffle_x86::forward(const Mat& bottom_blob, Mat& top_blob, const Optio
     const int outw = w * r;
     const int outh = h * r;
     const int outc = channels / (r * r);
-
-    if (bottom_blob.elempack != 1)
-        return PixelShuffle::forward(bottom_blob, top_blob, opt);
 
     if (r != 2 && r != 4)
         return PixelShuffle::forward(bottom_blob, top_blob, opt);
--- a/src/layer/x86/pixelshuffle_x86.cpp
+++ b/src/layer/x86/pixelshuffle_x86.cpp
@@ -1,0 +1,152 @@
+// Copyright 2026 Tencent
+// SPDX-License-Identifier: BSD-3-Clause
+
+#include "pixelshuffle_x86.h"
+
+#if __SSE2__
+#include <emmintrin.h>
+#if __AVX__
+#include <immintrin.h>
+#endif // __AVX__
+#endif // __SSE2__
+
+#include "x86_usability.h"
+
+namespace ncnn {
+
+PixelShuffle_x86::PixelShuffle_x86()
+{
+}
+
+int PixelShuffle_x86::forward(const Mat& bottom_blob, Mat& top_blob, const Option& opt) const
+{
+    int elembits = bottom_blob.elembits();
+    if (elembits != 32)
+        return PixelShuffle::forward(bottom_blob, top_blob, opt);
+
+    const int w = bottom_blob.w;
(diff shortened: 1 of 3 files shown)


nihui (approved):

nihui:

PR #6726, opened 2026-05-16
fix: avoid SIGFPE in fwrite_weight_data when Mat is empty

llx8:
## Problem

When running `ncnn2int8` to quantize YOLOv8 models, the tool crashes with `SIGFPE` (Arithmetic exception) during the `save()` phase.

The crash occurs in `ModelWriter::fwrite_weight_data()` in `tools/modelwriter.h`:

```cpp
ncnn::Mat data_flattened = data.reshape(data.w * data.h * data.d * data.c);
```

When `data` is an empty Mat (e.g. for certain layers in YOLOv8 detection head), `data.w * data.h * data.d * data.c` evaluates to 0, causing a division by zero inside `Mat::reshape()`.

## Fix

Add a null check before calling reshape:

```cpp
if (!data.data || data.w * data.h * data.d * data.c == 0) return 0;
```

## Reproduction

1. Convert YOLOv8n to ncnn format
2. Run `ncnn2int8` with a calibration table
3. Tool crashes with SIGFPE after quantizing all convolution layers


tencent-adm:
[CLA assistant check](https://cla.opensource.tencent.com/Tencent/ncnn?pullRequest=6726) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you all sign our [Contributor License Agreement](https://cla.opensource.tencent.com/Tencent/ncnn?pullRequest=6726) before we can accept your contribution.<br/>**1** out of **2** committers have signed the CLA.<br/><br/>:white_check_mark: lx-99-lyt<br/>:x: nihui<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.opensource.tencent.com/check/Tencent/ncnn?pullRequest=6726) it.</sub>

nihui:
Thanks for tracking this down and sending the fix.

I’ve updated the patch to handle this at the ModelWriter call sites instead of making `fwrite_weight_data()` silently skip empty `Mat`s. This keeps optional weights aligned with each layer’s `load_model()` logic, while still
preserving errors for required missing weights.

Thanks again for the contribution.

PR #6768, opened 2026-06-03
arm: add RotaryEmbed bf16 NEON support

devin-lai:
This is a follow-up to my previous RotaryEmbed bf16 PR, with an extra optimization pass based on @nihui's feedback. Thanks to @nihui for the guidance here.

After benchmarking the first version, I found that the non-interleaved bf16 path was already faster than fp32, but the interleaved path was still around break-even and could be slower on smaller ed values. The issue was that the bf16 interleaved path was still using a 4-wide 64-bit load/store pattern, while the fp32 path was already using full 128-bit NEON loads.

This update widens the bf16 and fp16-storage paths to process 8 lanes at a time with 128-bit NEON loads/stores. The math is still done in fp32, so the output stays unchanged.

The results are better now:

* interleaved ed=64: 0.84x -> about 1.30x vs fp32
* large interleaved case: about 15% faster than the previous bf16 version
* non-interleaved cases improved as well

I also verified the 8-wide, 4-wide, and scalar fallback paths. `test_rotaryembed` passes with `NCNN_BF16=ON`, ASan, and `NCNN_BF16=OFF`.

### DIFF
--- a/src/layer/arm/rotaryembed_arm.cpp
+++ b/src/layer/arm/rotaryembed_arm.cpp
@@ -7,6 +7,7 @@
 #include <arm_neon.h>
 #endif // __ARM_NEON
 
+#include "arm_usability.h"
 #include "cpu.h"
 
 namespace ncnn {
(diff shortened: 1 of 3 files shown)


nihui (approved):

nihui:

PR #6776, opened 2026-06-09
WIP: Persistent kvcache

MollySophia:
## Summary

This PR adds `SDPA` `kv_cache=2`, a persistent K/V cache mode for CPU.

`kv_cache=2` takes preallocated `past_key` / `past_value` cache buffers as
views. The input view height is the current valid cache length, while `cstep`
keeps the preallocated cache capacity. SDPA appends current K/V in place and
returns views of the same cache buffers with the updated valid length.

This avoids changing cache ownership every decode step and provides the base
interface for follow-up optimized kernels such as arm64/Vulkan flash attention.

Existing `kv_cache=0` and `kv_cache=1` behavior is kept unchanged.

Vulkan `kv_cache=2` is disabled in this PR. The existing Vulkan SDPA path still
uses the old concat-cache layout and cannot interpret the persistent cache view
layout correctly. Vulkan persistent-cache support will be handled in a follow-up
flash-attention PR.

## Changes

- Add `kv_cache=2` validation and input parsing in generic `SDPA`
- Add CPU implementations for generic, arm, x86, mips, and loongarch paths
- Append `cur_key` / `cur_value` into preallocated cache buffers at the input view height
- Return persistent cache buffers as updated views in `top_blobs[1]` and `top_blobs[2]`
- Add `test_sdpa_kvcache` coverage for normal use, invalid views, and tail dimensions
- Add a direct persistent-buffer semantic test for buffer identity, append correctness, and capacity overflow
- Disable Vulkan SDPA for `kv_cache=2` until the Vulkan path implements the persistent layout
- Route arm `kv_cache=2` int8 SDPA through the generic int8 implementation for correctness

The new code is only selected for `kv_cache=2`. Existing `kv_cache!=2` cases
continue to use the previous concat / no-cache paths.

`kv_cache=2` input layout:

```text
query, cur_key, cur_value, [attn_mask], past_key_view, past_value_view
```

`past_key_view.h` and `past_value_view.h` are the current valid cache
length. The underlying `cstep` keeps the cache capacity.

Output layout:

```text
out, past_key_buffer, past_value_buffer
```

## Correctness

Tested on aarch64:

```sh
cmake --build build-perf-linux --target test_sdpa test_sdpa_kvcache
taskset -c 0 ./build-perf-linux/tests/test_sdpa
taskset -c 0 ./build-perf-linux/tests/test_sdpa_kvcache
```

Both tests passed. The `test_sdpa_kvcache` coverage compares `kv_cache=2`
attention output numerically against reference output, including the int8 path,
and checks persistent cache buffer identity / view shape separately.

## Performance

Baseline:

```text
bedf9c65681a0b1e012081d43a566d20aaadf5c6
```

Current:

```text
cd3c5d44
```

Measured on Qualcomm Oryon aarch64, CPU0-3 governor `performance`.
Single-thread runs use `taskset -c 0`; four-thread runs use `taskset -c 0-3`.

Shape:

```text
d_k=192 d_v=128 heads=128 groups=16 dtype=fp16 ctx=4096
```

This PR is mainly an interface and cache-lifetime change. It avoids allocating a
new returned K/V cache blob every decode step: the output cache blobs are shallow
views of the caller-owned persistent cache buffers with updated `h`.

The current CPU fallback still builds compact fp32 K/V temporaries for the
existing Gemm-based SDPA math path, so latency is roughly neutral and temporary
per-call memory usage does not improve much by itself. The persistent view
layout is intended to be used by follow-up optimized kernels that can read
directly from the stable cache buffer.

### Causal prefill latency

| len | baseline t=1 ms | kv_cache=2 t=1 ms | baseline t=4 ms | kv_cache=2 t=4 ms |
|---:|---:|---:|---:|---:|
| 128 | 15.3350 | 15.2490 | 3.9160 | 4.0598 |
| 256 | 59.5649 | 59.7761 | 14.7642 | 14.9651 |
| 512 | 239.1162 | 239.2300 | 59.6160 | 60.3730 |
| 1024 | 944.7761 | 936.3608 | 248.9351 | 249.1870 |

### Decode latency

| past | baseline t=1 ms | kv_cache=2 t=1 ms | baseline t=4 ms | kv_cache=2 t=4 ms |
|---:|---:|---:|---:|---:|
| 128 | 0.7490 | 0.7329 | 0.1951 | 0.1941 |
| 512 | 2.2429 | 2.3611 | 0.7478 | 0.7119 |
| 1024 | 4.7610 | 4.5950 | 1.5261 | 1.5559 |
| 2048 | 10.1528 | 9.3242 | 3.5171 | 3.1182 |

### Decode temporary peak memory

Per-call temporary allocator peak, four threads:

```text
op_peak = workspace peak + blob peak
```

This only counts allocations made during one SDPA call after inputs are already
prepared. It does not include the caller-owned persistent K/V cache buffer.
In the current fallback path, SDPA still copies the valid cache range into
compact K/V blobs for the existing Gemm implementation, so the temporary peak
remains close to `kv_cache=1`.

| past | baseline MiB | kv_cache=2 MiB |
|---:|---:|---:|
| 128 | 1.69 | 1.74 |
| 512 | 6.22 | 6.46 |
| 1024 | 13.24 | 12.21 |
| 2048 | 23.58 | 25.87 |

For this shape, the persistent K/V cache buffer for `ctx=4096` is about 40 MiB:

```text
(d_k + d_v) * groups * sizeof(fp16) * ctx
= (192 + 128) * 16 * 2 * 4096
= 40 MiB
```

That buffer is allocated and owned outside the SDPA call, then reused across
decode steps. It replaces the growing `kv_cache=1` returned-cache ownership
model, but it is not shown in the temporary peak table above.


tencent-adm:
[CLA assistant check](https://cla.opensource.tencent.com/Tencent/ncnn?pullRequest=6776) <br/>All committers have signed the CLA.

nihui:
superseded by https://github.com/Tencent/ncnn/pull/6901

PR #6777, opened 2026-06-13
Fix int8 Convolution for flattened 1D input + 1x1 kernel

futz12:
The fp32 path of Convolution detects when bottom_blob.dims == 1 and kernel_w == kernel_h == 1 and redirects the computation to InnerProduct. However, the int8 inference paths (Convolution::forward_int8 and the architecture-specific forward_int8_*) did not have this fallback.

Expected broken behavior before this patch:
- For a flattened input with num_input elements, bottom_blob.dims == 1 and bottom_blob.c == 1, but the actual number of input channels is num_input.
- The int8 convolution loop iterated only over channels == 1, so most of the input values were ignored.
- The output shape was wrong: it became dims=3 with w=num_input, h=1, c=num_output instead of the correct dims=1, w=num_output.

This patch adds the missing flattened-blob handling:

- In the generic naive path (Convolution::forward_int8), dispatch to InnerProduct, mirroring the fp32 behavior.

- In the optimized x86 / arm / mips / loongarch paths (forward_int8_x86 / forward_int8_arm / forward_int8_mips / forward_int8_loongarch), reshape the 1D blob to 3D (1x1xW), recurse into the same int8 forward function, and reshape the result back to 1D.

A regression test tests/test_convolution_5.cpp is added. It compares the fp32 reference output (via InnerProduct) against the int8 output for dims==1 inputs with 1x1 kernels. Before the fix the test fails because the int8 path produced an incorrect shape; after the fix it passes with max abs error below 0.01.

### DIFF
--- a/src/layer/arm/convolution_arm.cpp
+++ b/src/layer/arm/convolution_arm.cpp
@@ -1299,6 +1299,25 @@ int Convolution_arm::forward_int8_arm(const Mat& bottom_blob, Mat& top_blob, con
 {
     int elembits = bottom_blob.elembits();
 
+    // flattened blob, implement as InnerProduct
+    if (bottom_blob.dims == 1 && kernel_w == 1 && kernel_h == 1)
+    {
+        Mat bottom_blob_3d = bottom_blob.reshape(1, 1, bottom_blob.w, opt.workspace_allocator);
+        if (bottom_blob_3d.empty())
+            return -100;
+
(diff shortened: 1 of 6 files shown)


nihui:
Convolution accepting 1d input is a deprecated behavior. In https://github.com/Tencent/ncnn/pull/6795, pnnx automatically converts the corresponding convolution to InnerProduct.

PR #6801, opened 2026-07-04
[perf] vulkan: SDPA KV Cache update

crafcat7:
## Summary

This PR updates the Vulkan `SDPA` KV cache path so K/V cache materialization is handled inside `SDPA_vulkan` with dedicated Vulkan shaders instead of a generic cache concat path.

The final implementation keeps public K/V cache outputs compact. This is required because `VkMat` is the shared Vulkan tensor representation used by ncnn layers, packing conversion, download/extract, and tests. A cache tensor returned from `SDPA_vulkan::forward()` must therefore expose a normal, self-consistent `w/h/c/cstep` layout.

Committed code changes are limited to Vulkan SDPA implementation files and shaders. This PR does not modify benchmark tools or external tooling.

### 1. Dedicated KV cache update shaders

The Vulkan SDPA KV cache path now uses SDPA-specific shaders for cache materialization:

```text
past_key   + cur_key   -> new_key_cache
past_value + cur_value -> new_value_cache
```

The new shader roles are:

- `sdpa_kvcache_update.comp`
  - Copies non-empty `past` and current `cur` into a compact `past + cur` cache output.
- `sdpa_kvcache_append.comp`
  - Copies current `cur` into the output at `past_seqlen + y`.
  - Used by the compact-output path for empty-past initialization.

This removes the dependency on a generic concat layer for the SDPA KV cache update step and makes the cache path explicit inside `SDPA_vulkan`.

### 2. Compact public cache outputs

The final cache output layout is:

```text
w = head_dim
h = past_seqlen + cur_seqlen
c = kv_group_count
cstep ~= w * h
```

The public `VkMat` output does not encode hidden cache capacity in `cstep`.

### 3. Existing attention compute paths remain unchanged

The cache update happens before attention kernels read K/V. The existing SDPA compute paths are preserved:

- flash attention
- cooperative-matrix flash attention
- regular `qk -> softmax -> qkv`

### Comparison basis

All results compare clean worktrees built with the same compiler, CMake options, and benchmark parameters:

- Baseline: `81e8cbff`
- Optimized: `39cc7404`
- GPU: NVIDIA GeForce RTX 5070 Ti
- Build: Windows/MSVC Release, Vulkan enabled
- CPU side: scalar x86 options in this local build to avoid a local `Illegal instruction` issue
- Run order: baseline, optimized, optimized, baseline

Improvement is calculated as:

```text
(baseline time - optimized time) / baseline time * 100%
```

### End-to-end LLM benchmark

The normal committed benchmark path was used. It keeps K/V cache as host `Mat` between decoder calls; no GPU-resident cache mode or modified benchmark path is involved.

```text
benchncnn_llm.exe 3 1 0 0 0
```

Parameters and sampling:

- `loop_count = 3`
- `num_threads = 1`
- `powersave = 0`
- `gpu_device = 0`
- `cooling_down = 0`
- 8 warmup iterations per case, as defined by the committed benchmark
- two complete runs per revision; table values are the mean of the two reported averages
- decode case: prefill 256 tokens, then decode one token with a 256-token KV cache

| Model | Baseline decode ms | Optimized decode ms | Improvement |
| --- | ---: | ---: | ---: |
| hunyuan_0.5b | 571.27 | 576.56 | -0.93% |
| minicpm4_0.5b | 239.00 | 231.84 | +2.99% |
| qwen2.5_0.5b | 49.31 | 52.45 | -6.36% |
| qwen3_0.6b | 670.12 | 664.28 | +0.87% |
| llama3.2_1b | 377.49 | 378.88 | -0.37% |
| tinyllama_1.1b | 527.79 | 524.04 | +0.71% |
| youtu_llm_2b | 986.72 | 971.26 | +1.57% |
| Suite total | 3421.69 | 3399.29 | **+0.65%** |

The prefill suite total changed from `3330.81 ms` to `3384.11 ms` (`-1.60%`). Prefill has no existing cache to concatenate, so it is a control case rather than the target of this optimization.

The end-to-end result is small and noisy because model compute and host `Mat` cache upload/download dominate the dedicated KV update dispatch. It is still the comparable normal benchmark result: across the complete decode suite, the optimized revision reduces measured time by `0.65%`.

### Isolated Vulkan SDPA decode benchmark

`perf_sdpa_decode` isolates `SDPA` forward and uploads inputs before timing, so it exposes the cost removed by replacing generic `Concat` with the SDPA-specific KV update pipeline.

Both revisions used the same measurement cases on `gpu-0`:

```text
query       = [128, 1, 16]
current K/V = [128, 1, 4]
past K/V    = [128, past_seqlen, 4]
past_seqlen = 128, 256, 512, 1024, 2048
```

Each value below is the mean of two runs. Each run uses the median of 20 timed command-buffer batches, divided by the calibrated batch size to obtain per-forward latency.

| Past tokens | FP32 baseline us | FP32 optimized us | FP32 improvement | FP16 arithmetic baseline us | FP16 arithmetic optimized us | FP16 improvement |
| ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| 128 | 34.56 | 30.85 | **+10.73%** | 101.52 | 99.00 | +2.48% |
| 256 | 50.23 | 46.54 | **+7.36%** | 176.35 | 174.11 | +1.27% |
| 512 | 81.45 | 76.15 | **+6.51%** | 326.25 | 325.35 | +0.28% |
| 1024 | 144.35 | 136.00 | **+5.78%** | 619.75 | 619.65 | +0.02% |
| 2048 | 273.65 | 259.25 | **+5.26%** | 1206.90 | 1208.55 | -0.14% |

The dedicated KV update path improves isolated FP32 SDPA decode by `5.26%` to `10.73%`. The FP16 arithmetic (`fp16psa`) path improves by up to `2.48%` at shorter cache lengths and becomes neutral as attention computation dominates at longer lengths.

These two benchmark levels describe different effects:

```text
generic Concat -> dedicated KV update shader
    |
    +-> isolated SDPA: visible dispatch and cache-update reduction
    |
    +-> complete decoder: mostly hidden by attention, GEMM, and host cache transfer
```


nihui:
superseded by https://github.com/Tencent/ncnn/pull/6901

PR #6815, opened 2026-07-10
optimize multi-head attention (increased shared-memory tile size, add fused kernel)

MaxwellGengYF:
Design from https://github.com/LuisaGroup/LuisaCompute/blob/next/examples/ml/multi_head_attention.cpp.
reviewed by human, not only AI Agent.


tencent-adm:
[CLA assistant check](https://cla.opensource.tencent.com/Tencent/ncnn?pullRequest=6815) <br/>All committers have signed the CLA.

nihui:
delete *.opt_bak files

MaxwellGengYF:
*.opt_bak deleted

nihui:
## Full review summary

Reviewed all 14 changed files at commit `5ee1210e236e4030f15e09b752c3a87d8102238a` line by line, including the Vulkan dispatch code, shader layouts/macros, benchmark, and build integration.

### Existing inline findings confirmed

1. The 16x16 pack4 shaders perform shared-memory source loads before the padded `gx/gy` bounds check. Rounded dispatches therefore read out of bounds for tail tiles; the benchmark's own `seq_len=16` optimized cases would trigger this once it actually uses Vulkan.
2. A 16x16 workgroup requires 256 invocations, while Vulkan Core only guarantees 128. `Pipeline::set_local_size_xyz` does not clamp against `maxComputeWorkGroupInvocations`, and both affected `Pipeline::create` return values are ignored.
3. `bench_mha_compare` never executes Vulkan. It creates the CPU layer and calls the `std::vector<Mat>` forward overload, so both sections labeled GPU time CPU execution.

### Additional blocking findings

1. **The enlarged QK tile exceeds the shared-memory limit of valid Vulkan devices.**

   [`multiheadattention_qk_cross_pack4.comp`](https://github.com/Tencent/ncnn/blob/5ee1210e236e4030f15e09b752c3a87d8102238a/src/layer/vulkan/shader/multiheadattention_qk_cross_pack4.comp#L30-L32) allocates 1,280 `lfpvec4` values. In pure FP32 this is 20,480 bytes, above Vulkan Core's guaranteed 16,384-byte `maxComputeSharedMemorySize`. Pipeline creation is not gated on `vkdev->info.max_shared_memory_size()` and its error is ignored, so valid 16-KiB devices can later record an invalid pipeline.

2. **The advertised subgroup and fused optimizations have no execution path.**

   The subgroup pipeline is [created here](https://github.com/Tencent/ncnn/blob/5ee1210e236e4030f15e09b752c3a87d8102238a/src/layer/vulkan/multiheadattention_vulkan.cpp#L190-L201), but [`forward` only selects the four legacy QK pipelines](https://github.com/Tencent/ncnn/blob/5ee1210e236e4030f15e09b752c3a87d8102238a/src/layer/vulkan/multiheadattention_vulkan.cpp#L510-L528). The four fused pipelines are never created or dispatched. The comment saying `NCNN_ENABLE_FUSED_MHA=1` enables them is inaccurate; no such option or environment lookup exists. Currently this adds shader registry/binary content and unnecessary subgroup pipeline compilation without providing the claimed optimization.

3. **The subgroup QK shader is logically incorrect if it is connected.**

   In [`multiheadattention_qk_subgroup.comp`](https://github.com/Tencent/ncnn/blob/5ee1210e236e4030f15e09b752c3a87d8102238a/src/layer/vulkan/shader/multiheadattention_qk_subgroup.comp#L38-L95), each subgroup lane has a different `(gx, gy)`, but `subgroupAdd` combines products from those unrelated output coordinates and only lane zero writes. A 64-lane subgroup therefore writes one of 64 expected scores, containing a mixture of different rows/columns. Its non-3D mask branch also discards the supplied 2-D additive mask and invents a causal mask.

   Pipeline creation additionally ignores `opt.use_subgroup_ops`. FP16 arithmetic needs `GL_EXT_shader_subgroup_extended_types_float16`, which is neither requested nor feature-gated, so this shader fails compilation for valid option/device combinations.

4. **All four fused shaders need substantial correction before they can be enabled.**

   - Three declare `attn_mask_blob` as `sfpvec4[]` and then use scalar `buffer_ld1`; focused glslang checks reproduce vec4-to-scalar compilation failures.
   - The cooperative-matrix shader requires SPIR-V 1.6, while ncnn's cooperative shader path targets SPIR-V 1.3.
   - Their Q/K/V/output indexing assumes row-major `[head][position][dimension]`, but the actual affine `VkMat` layout is sequence-width and embedding-height.
   - Q projection already applies the layer's attention scale; all four multiply by `1/sqrt(head_dim)` again.
   - They reinterpret every 2-D additive mask as a causal mask.
   - The regular, coopload and subgroup variants retain/write only eight vec4 accumulators (32 channels), so typical `head_dim=64` leaves half the result unwritten.
   - The coopload variant returns lanes before barriers, shares lane zero's head data with workgroups that can span several heads, and lacks a barrier before shared memory is overwritten.
   - The `coopmat` variant never uses a cooperative matrix and indexes `float o_accum[64]` out of bounds for `head_dim > 64`.

5. **The benchmark would still not isolate this optimization after fixing GPU execution.**

   [`opt_default` disables packing while `opt_optimized` enables it](https://github.com/Tencent/ncnn/blob/5ee1210e236e4030f15e09b752c3a87d8102238a/benchmark/bench_mha_compare.cpp#L139-L183), so the result conflates pack1-vs-pack4 with 8x8-vs-16x16 tiles. Every call also regenerates different inputs and weights, ignores forward return codes, and never compares outputs. A useful benchmark should reuse identical uploaded tensors and compare old/new pack4 kernels with correctness validation and synchronized GPU timing.

6. **Build/performance integration issues.**

   - [`bench_mha_compare` is added unconditionally](https://github.com/Tencent/ncnn/blob/5ee1210e236e4030f15e09b752c3a87d8102238a/benchmark/CMakeLists.txt#L72-L74). CPU-only builds produce an executable that can only fail, and disabling the MultiHeadAttention layer makes `create_layer` return null and the benchmark dereference it.
   - The 16x16 local size is also selected when shader-local memory is disabled. That branch has no tile reuse and can schedule 256 invocations for an 8x8 useful region instead of the previous 64.
   - `loop_count=0` is accepted and produces `0/0` average timing while printing `DBL_MAX` as the minimum.

### Validation

- Vulkan Release build including `bench_mha_compare`: passed.
- Existing `test_multiheadattention` GPU test on llvmpipe: passed.
- Focused GLSL checks reproduced the dormant subgroup/fused compilation failures.
- `git diff --check`: clean.

The passing llvmpipe test does not cover 128-invocation/16-KiB devices, and none of the new fused/subgroup kernels is currently dispatched. I recommend addressing the active pack4 safety/portability regressions and removing or completing the dormant kernels before merging.

PR #6834, opened 2026-07-17
RotaryEmbed: support full-width cos/sin cache for 2D / vision RoPE

### LINKED ISSUE #6833: RotaryEmbed  drops the second half of cos/sin — wrong output for 2D / vision RoPE
## Summary

`RotaryEmbed`'s non-interleaved path reads only the first `embed_dim/2` values of the
`cos`/`sin` cache and reuses them for **both** halves of each head vector:

```cpp
// src/layer/rotaryembed.cpp (naive), same in arm/x86/vulkan/loongarch/mips
for (int j = 0; j < embed_dim / 2; j++)
{
    const float cos_val = *cos_ptr++;   // only reads cos[0 .. embed_dim/2)
    const float sin_val = *sin_ptr++;
    *outptr0++ = x0 * cos_val - x1 * sin_val;   // out[j]
    *outptr1++ = x0 * sin_val + x1 * cos_val;   // out[j+half] reuses the SAME cos/sin
}
```

This is correct for standard 1-D RoPE, where `cos = concat(freqs, freqs)` and the two
halves are identical. But `fuse_convert_rotaryembed` lowers the general HF expression
`x*cos + rotate_half(x)*sin` (with a **full-width** `cos`) into this op **without checking
that the two halves are equal**. For models whose halves differ — 2-D / vision RoPE, where
`cos = concat(cos_h, cos_w)` — the second output half is rotated with the wrong (height)
frequencies instead of the width frequencies, silently producing wrong output. 

The correct general form (directly matching the fused expression) is:

```
out[j]      = x[j]*cos[j]        - x[j+half]*sin[j]
out[j+half] = x[j+half]*cos[j+half] + x[j]*sin[j+half]   // <-- second-half cos/sin
```

Found this  in trying #6788.

## Reproduction

Vision encoder of `tencent/Penguin-VL-2B` (Qwen3-0.6B-based, bidirectional, 2-D RoPE;
`cos = concat(cos_h, cos_w)`, the two 64-wide halves differ by up to 2.0). Converted with
pnnx, run in fp32 blobs, compared against the PyTorch reference encoder output (1024 tokens
× 1024 dims):

| ncnn `RotaryEmbed` | `max-diff` vs PyTorch | per-token cosine sim (mean) |
| --- | --- | --- |
| current (reads half) | **3.39** | **0.45** |
| patched (reads full width) | **6.2e-4** | ~1.0 |

The 3.39 is not a precision artifact — it is deterministic (identical for fp16 and fp32
weights) and catastrophic (mean per-token cosine similarity 0.45; all 1024 tokens have >10% relative error). The PyTorch-side wrapper matches the reference encoder to 5.7e-5, so
the divergence is entirely in the op. It is masked in end-to-end tests because greedy
`argmax` decoding tolerates a badly wrong vision encoding for simple inputs.

## Proposed fix

Read the second-half `cos`/`sin` for the second output half. Backward-compatible:

- For standard RoPE the two halves are identical, so the result is **bit-identical**
  (verified: patched vs current build byte-for-byte on the same graph with identical-halves
  cos).
- Guard on `cos_cache.w`: use the second half only when the cache is full width
  (`w == embed_dim`); when it is `embed_dim/2` (as `test_rotaryembed.cpp` supplies today),
  fall back to the current duplicated behavior so existing half-width callers keep working.
- Add a differing-halves test case so the path is covered.

Happy to send a PR covering all backends (naive / arm / x86 / vulkan / loongarch / mips)
plus the test. This unblocks in-graph conversion of 2-D vision RoPE (Penguin-VL, and any
`Qwen-VL`-style vision encoder that concatenates per-axis frequencies), which currently
forces ports to compute RoPE outside the graph.

Acture:
## What

`RotaryEmbed`'s non-interleaved path read only the first `embed_dim/2` values of the cos/sin
cache and reused them for **both** halves of each head vector. That is correct for standard 1-D
RoPE (`cos = concat(freqs, freqs)`, identical halves), but `fuse_convert_rotaryembed` lowers the
general expression `x*cos + rotate_half(x)*sin` — with a **full-width** cos — into this op without
checking that the halves are equal. For 2-D / vision RoPE where the two halves differ
(`cos = concat(cos_h, cos_w)`, e.g. the Penguin-VL / Qwen-VL-style vision encoder) the second half
was rotated with the wrong frequencies, silently producing wrong output.

Fixes #6833.

## Fix

Read the second half's cos/sin for the second output half. Backward-compatible via a
cache-width guard:

- full-width cache (`cos_cache.w == embed_dim`) → the two halves are independent (2-D rope);
- half-width cache (`embed_dim/2`, the standard-rope form the unit test supplies) → the second
  half reuses the first, so existing half-width callers are **bit-identical**.

Applied to every backend (naive / arm / x86 / loongarch / mips / vulkan) and their bf16/fp16 paths.
The vulkan shaders take the cache row width as a new push constant.

## Test

`tests/test_rotaryembed.cpp` previously only used a half-width cache, which is trivially
identical-halves and never exercised the differing-halves path. Added a full-width
differing-halves case for the non-interleaved op.

## Verification

- `test_rotaryembed` and `test_rotaryembed_oom` pass locally on the **naive + ARM** backends
  (macOS Apple Silicon), covering both the existing half-width and the new full-width cases.
- End-to-end: the Penguin-VL-2B vision encoder (28-layer bidirectional, 2-D RoPE) converted with
  pnnx now runs in-graph at **max|diff| = 6.2e-4** vs the PyTorch reference (fp32); before the fix
  the same fused graph gave **max|diff| = 3.39**.
- Standard-rope regression: patched vs unpatched build produce **byte-identical** output on a
  fused graph fed an identical-halves cache.
- x86 / loongarch / mips / vulkan carry the same mechanical change but were not built locally
  (no toolchain on this host) — relying on CI to validate them.


tencent-adm:
[CLA assistant check](https://cla.opensource.tencent.com/Tencent/ncnn?pullRequest=6834) <br/>All committers have signed the CLA.

Acture:
> [CLA assistant check](https://cla.opensource.tencent.com/Tencent/ncnn?pullRequest=6834) Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.opensource.tencent.com/Tencent/ncnn?pullRequest=6834) before we can accept your contribution.You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.opensource.tencent.com/check/Tencent/ncnn?pullRequest=6834) it.

~~Got Internal Server Error when I tried to sign this.~~

Signed.

Acture:
@codex review

Acture:
@dosu any concerns with this fix?

Issue #3555, opened 2022-02-10
Incorrect scaling in Interp layer coming from ONNX


ghost:
## context | 编译/运行环境 | バックグラウンド
Running a PyTorch model 
1. Exported to ONNX
2. Simplified through onnxsim using args `--dynamic-input-shape --input-shape input:1,1,86,86`
3. Converted to ncnn using onnx2ncnn.

## how to reproduce | 复现步骤 | 再現方法
1. Run the attached param file with input shape (86, 86, 1) 
2. Observe that output shape is (255, 255, 1) whereas the correct shape is (256, 256, 1)
3. The issue comes from the rounding rules in for ONNX operator  https://github.com/onnx/onnx/blob/main/docs/Changelog.md#Resize-11
[interp onnx scale.param.txt](https://github.com/Tencent/ncnn/files/8041239/interp.onnx.scale.param.txt)

## Diff for fix
```
diff --git a/src/layer/interp.cpp b/src/layer/interp.cpp
index f6df94c6..16e329db 100644
--- a/src/layer/interp.cpp
+++ b/src/layer/interp.cpp
@@ -437,8 +437,9 @@ int Interp::forward(const Mat& bottom_blob, Mat& top_blob, const Option& opt) co
     }
     if (outw == 0 || outh == 0)
     {
-        outw = static_cast<int>(w * width_scale);
-        outh = static_cast<int>(h * height_scale);
+        // Calle: assume half_pixel ONNX rounding mode
+        outw = static_cast<int>((w + 0.5f) * width_scale - 0.5f);
+        outh = static_cast<int>((h + 0.5f) * height_scale - 0.5f);
     }
 
     Mat reference_blob;
****
```

nihui:
confirmed

nihui:
针对onnx模型转换的各种问题，推荐使用最新的pnnx工具转换到ncnn
In view of various problems in onnx model conversion, it is recommended to use the latest pnnx tool to convert your model to ncnn

```shell
pip install pnnx
pnnx model.onnx inputshape=[1,3,224,224]
```

详细参考文档
Detailed reference documentation
https://github.com/pnnx/pnnx
https://github.com/Tencent/ncnn/wiki/use-ncnn-with-pytorch-or-onnx#how-to-use-pnnx

PR #6837, opened 2026-07-20
Interp: fix bilinear/bicubic coordinate scale for non-integer scale_factor (same as #3555)

### LINKED ISSUE #6836: Interp bilinear/bicubic use the wrong sampling scale for non-integer scale_factor (same as #3555, but for bilinear/bicubic)
Hit this converting a model with `F.interpolate(x, scale_factor=1.5, mode='bilinear')` — ncnn doesn't match PyTorch, but only for bilinear/bicubic. nearest is fine.

It's the same problem as #3555: when resizing by a scale factor (no explicit size), the coordinate scale should be `1/scale_factor`, not `w/outw` (input over rounded output). #3555 / 76e32e9 fixed that for **nearest**, but bilinear/bicubic still compute `w/outw` in `linear_coeffs`/`cubic_coeffs`. For integer scales the two are equal so it never shows up; for 1.5x on 7px, outw=10 and `7/10=0.7` vs `1/1.5=0.667` — every sample is offset.

Affects both front ends (PyTorch's default `recompute_scale_factor=False`, and ONNX Resize `half_pixel` — same as #3555).

Repro (pnnx, 3x7x7 fp32, vs PyTorch):

```
bilinear  scale=1.5  ac=False   max|diff| = 0.28
bicubic   scale=1.5  ac=False   max|diff| = 0.36
bilinear  scale=2.0  ac=False   0    (integer)
bilinear  scale=1.5  ac=True    0
nearest   scale=1.5             0
bilinear  size=(10,10)          0    (explicit size)
```

So it's bilinear/bicubic + non-integer scale + align_corners=False, no warning during conversion.

Fix is the same shape as 76e32e9 — use `1/width_scale` in the scale-factor path instead of `w/outw`. Explicit size stays `w/outw` (correct), integer scales stay bit-exact. I've got it working locally across the backends + a test; happy to send a PR.

Acture:
## What

When `Interp` resizes by a scale factor (no explicit output size), the coordinate scale should be `1/scale_factor`, not `w/outw` (input over the *rounded* output). #3555 / 76e32e9 fixed this for **nearest**, but bilinear and bicubic still used `w/outw` in `linear_coeffs`/`cubic_coeffs`.

For integer scales the two are equal, so it never showed up. For a non-integer scale (1.5x on 7 → outw=10, `7/10=0.7` vs `1/1.5=0.667`) every sample is offset — matching neither PyTorch's default (`recompute_scale_factor=False`) nor ONNX Resize `half_pixel`. No warning during conversion.

Fixes #6836.

## Repro (pnnx, 3x7x7 fp32, vs PyTorch — before this PR)

```
bilinear  scale=1.5  ac=False   max|diff| = 0.28
bicubic   scale=1.5  ac=False   max|diff| = 0.36
bilinear  scale=2.0  ac=False   0    (integer)
bilinear  scale=1.5  ac=True    0
nearest   scale=1.5             0
bilinear  size=(10,10)          0    (explicit size)
```

## Fix

`linear_coeffs`/`cubic_coeffs` now take the coordinate scale as an argument (with a thin wrapper keeping old callers). Each call site passes `1/width_scale` / `1/height_scale` in the scale-factor path and `w/outw` in the explicit-size path — same as the nearest path already does. `align_corner` and explicit-size are untouched, and integer scales stay bit-exact.

Applied across every backend (naive / arm / x86 / loongarch / mips / riscv incl. bf16/fp16 paths) and vulkan.

## Verification

- End-to-end vs PyTorch on the **naive and ARM** backends: the two triggering cases go `0.28 / 0.36 → 0`, controls stay 0.
- `test_interp` passes locally (its existing non-integer bilinear/bicubic cases compare each backend against naive, so they guard cross-backend consistency).
- x86 / loongarch / mips / riscv / vulkan carry the same mechanical change but were not built locally (no toolchain/GPU on this host) — relying on CI.

### DIFF
--- a/src/layer/arm/interp_arm.cpp
+++ b/src/layer/arm/interp_arm.cpp
@@ -135,7 +135,7 @@ int Interp_arm::forward(const std::vector<Mat>& bottom_blobs, std::vector<Mat>& 
         {
             if (resize_type == 1) // nearest
             {
-                const float ws = (output_width || !size_expr.empty()) ? w / (float)outw : 1.f / width_scale;
+                const float ws = (output_width || dynamic_target_size || !size_expr.empty()) ? w / (float)outw : 1.f / width_scale;
 
                 #pragma omp parallel for num_threads(opt.num_threads)
                 for (int y = 0; y < h; y++)
@@ -161,7 +161,7 @@ int Interp_arm::forward(const std::vector<Mat>& bottom_blobs, std::vector<Mat>& 
                 int* xofs = buf;
                 float* alpha = (float*)(buf + outw);
 
-                const float ws = (output_width || !size_expr.empty()) ? w / (float)outw : 1.f / width_scale;
+                const float ws = (output_width || dynamic_target_size || !size_expr.empty()) ? w / (float)outw : 1.f / width_scale;
                 linear_coeffs(w, outw, ws, xofs, alpha, align_corner);
 
                 #pragma omp parallel for num_threads(opt.num_threads)
@@ -199,7 +199,7 @@ int Interp_arm::forward(const std::vector<Mat>& bottom_blobs, std::vector<Mat>& 
                 int* xofs = buf;
                 float* alpha = (float*)(buf + outw);
 
-                const float ws = (output_width || !size_expr.empty()) ? w / (float)outw : 1.f / width_scale;
+                const float ws = (output_width || dynamic_target_size || !size_expr.empty()) ? w / (float)outw : 1.f / width_scale;
                 cubic_coeffs(w, outw, ws, xofs, alpha, align_corner);
 
                 #pragma omp parallel for num_threads(opt.num_threads)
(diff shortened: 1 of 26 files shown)


Acture:
Good catch — fixed in 57f739a. The condition now also checks \`dynamic_target_size\`, so the scale-factor branch (\`1/scale\`) is only taken for an actual scale factor; the reference-blob / explicit-size cases use \`w/outw\`. This also covers the same latent case in the nearest path. `test_interp` (which exercises the dynamic_target_size path via `test_interp_ref`) still passes.

@codex review

Acture:
Fixed in 9c69ad3 — good catch again. The bf16 helper now takes `dynamic_target_size` and checks `(output_width || dynamic_target_size || has_size_expr)` in every coeff site, matching the fp32 x86 path, so reference-blob resizes on the bf16 storage path use `w/outw`. `test_interp` passes. @codex review

PR #6851, opened 2026-07-26
fix(interp): use 1/scale_factor coordinate scale for bilinear/bicubic resize

### LINKED ISSUE #6836: Interp bilinear/bicubic use the wrong sampling scale for non-integer scale_factor (same as #3555, but for bilinear/bicubic)

Lfan-ke:
`Interp`'s `linear_coeffs`/`cubic_coeffs` always derive the sampling scale from `w / outw`, but when resizing by a scale factor (no explicit output size) the coordinate scale should be `1 / scale_factor`, matching PyTorch. #3555 fixed this for nearest; bilinear/bicubic were missed.

For integer scale factors the two are equal, so this only affects non-integer factors, e.g. `F.interpolate(x, scale_factor=1.5, mode='bilinear')`.

**Fix**: thread the scale-factor-aware scale (the same expression already used on the nearest path) into `linear_coeffs`/`cubic_coeffs` across all backends. The vulkan path drops the `resize_type == 2` special-case so bilinear uses the same scale as nearest - the shader already applies the scale generically.

**Verification**: for a 5-wide input at `scale_factor=1.5` (outw=7), bilinear output changes from `0, 0.5714, 1.2857, ...` to `0, 0.5, 1.1667, 1.8333, 2.5, 3.1667, 3.8333`, matching PyTorch's `(dx+0.5)/1.5 - 0.5` mapping exactly. Verified locally on x86 (native) and aarch64 (cross-compiled + qemu) that each backend's output matches the corrected naive reference via `test_interp`.

Closes #6836

### DIFF
--- a/src/layer/arm/interp_arm.cpp
+++ b/src/layer/arm/interp_arm.cpp
@@ -135,7 +135,7 @@ int Interp_arm::forward(const std::vector<Mat>& bottom_blobs, std::vector<Mat>& 
         {
             if (resize_type == 1) // nearest
             {
-                const float ws = (output_width || !size_expr.empty()) ? w / (float)outw : 1.f / width_scale;
+                const float ws = (output_width || dynamic_target_size || !size_expr.empty()) ? w / (float)outw : 1.f / width_scale;
 
                 #pragma omp parallel for num_threads(opt.num_threads)
                 for (int y = 0; y < h; y++)
@@ -161,7 +161,7 @@ int Interp_arm::forward(const std::vector<Mat>& bottom_blobs, std::vector<Mat>& 
                 int* xofs = buf;
                 float* alpha = (float*)(buf + outw);
 
-                linear_coeffs(w, outw, xofs, alpha, align_corner);
+                linear_coeffs(w, outw, xofs, alpha, align_corner, (output_width || dynamic_target_size || !size_expr.empty()) ? (double)w / outw : 1.0 / width_scale);
 
                 #pragma omp parallel for num_threads(opt.num_threads)
                 for (int y = 0; y < h; y++)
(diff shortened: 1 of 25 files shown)


Lfan-ke:
Both real, fixed. Added `dynamic_target_size` to the target-size condition at every call site (naive + all backends, threaded through the `interp_bf16s` free functions), and gated the Vulkan bicubic `constants[2]` the same way. Independent check: a dynamic 16->12 resize now spans the source (`0.167 ... 14.833`, max err 0.0 vs the bug's `0 ... 11`); `test_interp` passes, and scale-factor resizes still use `1/width_scale`.

Lfan-ke:
Closing in favor of #6837 - it fixes the same non-integer scale_factor issue and also covers the pnnx F_interpolate pass, so this one is redundant.

PR #6864, opened 2026-07-27
vulkan: route download through staging when blob memory is not HOST_CACHED (13x, fixes #6857)

### LINKED ISSUE #6857: Vulkan: record_download() bypasses the HOST_CACHED staging buffer, so downloads run at 0.33 GB/s (12.7x fix)
### Summary

`VkCompute::record_download()` skips the staging buffer whenever the blob allocator is mappable. On hardware where the blob buffer is `DEVICE_LOCAL | HOST_VISIBLE` but **not** `HOST_CACHED` — i.e. write-combined — this makes ncnn `memcpy` results out of uncached memory. Uploads stay fast; downloads collapse.

Measured on a Radeon 8060S (Strix Halo, RADV GFX1151, 256 GB/s unified LPDDR5X):

| | throughput |
|---|---|
| plain host `memcpy` (reference) | 22.3 GB/s |
| ncnn **upload** | 10.3 GB/s |
| ncnn **download** | **0.33 GB/s** |

A 31× asymmetry between upload and download on the same buffers.

### Cause

`src/command.cpp`, in `VkCompute::record_download()`:

```cpp
Option opt_staging = opt;
if (!opt_staging.blob_vkallocator->mappable)
{
    opt_staging.blob_vkallocator = opt.staging_vkallocator;
}
```

The staging allocator is the one that asks for `HOST_CACHED`:

```cpp
// VkStagingAllocator::fastMalloc
buffer_memory_type_index = vkdev->find_memory_index(
    memoryRequirements.memoryTypeBits,
    VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT | VK_MEMORY_PROPERTY_HOST_COHERENT_BIT,
    VK_MEMORY_PROPERTY_HOST_CACHED_BIT,          // preferred
    VK_MEMORY_PROPERTY_DEVICE_LOCAL_BIT);
```

but it is bypassed exactly when it is needed, because the blob buffer *is* mappable — just not cached. On this device:

| Memory type | DEVICE_LOCAL | HOST_VISIBLE | HOST_COHERENT | HOST_CACHED |
|---|---|---|---|---|
| 3, 4, 9 | yes | yes | yes | **no** ← blob allocator picks these |
| 5, 6, 10 | – | yes | yes | **YES** ← staging allocator wants these |

The integrated-GPU path in `VkBlobAllocator::fastMalloc` deliberately prefers unified memory:

```cpp
if (vkdev->info.type() == 1)
{
    // integrated gpu, prefer unified memory
    buffer_memory_type_index = vkdev->find_memory_index(..., VK_MEMORY_PROPERTY_DEVICE_LOCAL_BIT, VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT, 0);
```

which is a good choice for uploads and for device access — it is only the host **read** that suffers.

**This is not specific to integrated GPUs.** The discrete + resizable-BAR path a few lines below passes `HOST_CACHED` as *preferred-not*:

```cpp
if (vkdev->info.resizable_bar_enabled())
{
    buffer_memory_type_index = vkdev->find_memory_index(..., VK_MEMORY_PROPERTY_DEVICE_LOCAL_BIT, VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT, VK_MEMORY_PROPERTY_HOST_CACHED_BIT);
```

so a discrete GPU with ReBAR enabled should hit the same path.

### Suggested fix

Also stage the download when the blob memory is not itself cached:

```cpp
if (!opt_staging.blob_vkallocator->mappable || !opt_staging.blob_vkallocator->cached)
{
    opt_staging.blob_vkallocator = opt.staging_vkallocator;
}
```

with a `cached` flag added alongside the existing `mappable` / `coherent` on `VkAllocator`, set from a new `VulkanDevice::is_cached()` that mirrors `is_coherent()`. 28 lines in total across `allocator.h`, `allocator.cpp`, `gpu.h`, `gpu.cpp`, `command.cpp`.

### Effect

```
download bandwidth   0.33 -> 4.17 GB/s   (12.7x)
```

Measured end to end on a 22-model YOLO zoo (ultralytics exports, 640×640, 4 threads, median of 3 sessions). **Before the change three models were slower on the GPU than on the CPU; afterwards none are:**

| Model | before | after |
|---|---|---|
| FastSAM-s | 4.41× | **8.70×** |
| yolov8n | 2.43× | 5.40× |
| yolov8n-seg | 1.67× | 5.68× |
| yolo26n-sem | **0.20×** | 3.03× |
| yoloe-11s-seg-pf | **0.40×** | 3.09× |
| yolo12n | **0.37×** | 1.81× |

Median raw forward pass went from ~2× to ~4.8×; end-to-end from 2.26× to 4.04×.

Outputs are unchanged — a YOLOE segmentation model still matches its PyTorch reference at corr 1.000000 (CPU) and 0.999994 (Vulkan, fp16 arithmetic).

Some headroom remains: 4.17 GB/s is still well under the 22.3 GB/s host `memcpy` reference, so the path still costs an extra copy beyond what the hardware requires.

### Environment

ncnn `a4d2ea1` (2026-07-22), Vulkan SDK 1.4.350.1, RADV, Ubuntu 24.04, Radeon 8060S Graphics (GFX1151), `rebar=1`.

I am happy to open a PR with the patch if the approach looks right — including whether `cached` belongs on `VkAllocator` or is better queried per allocation.

giuliocorradi:
Fixes #6857. Opening as requested in https://github.com/Tencent/ncnn/issues/6857#issuecomment — using the `is_cached()` on `VkAllocator` approach you suggested.

## Problem

`VkCompute::record_download()` skips the staging buffer whenever the blob allocator is mappable:

```cpp
Option opt_staging = opt;
if (!opt_staging.blob_vkallocator->mappable)
{
    opt_staging.blob_vkallocator = opt.staging_vkallocator;
}
```

On an integrated GPU the blob allocator prefers `DEVICE_LOCAL | HOST_VISIBLE` unified memory. That memory *is* mappable — but it is **write-combined**, not `HOST_CACHED`. So the result gets `memcpy`'d out of uncached memory.

Uploads are unaffected (writing to write-combined memory is fast). Only the device-to-host read collapses, which is why the same buffers show a large upload/download asymmetry.

The staging allocator is precisely the one that requests `HOST_CACHED`, and it was bypassed in exactly the case where it is needed.

## Change

28 lines across 5 files:

- `VulkanDevice::is_cached()` — mirrors `is_coherent()`
- `VkAllocator::cached` — tracked alongside `mappable` / `coherent`, set at every site those two are set
- `record_download()` — requires mappable **and** cached to take the direct path

No behaviour change on discrete GPUs (blob memory is not mappable, so the staging path was already taken) or on hardware whose mappable blob memory *is* `HOST_CACHED` (condition still passes).

## Measurements

Radeon 8060S (Strix Halo, RADV GFX1151, 256 GB/s unified LPDDR5X), 128 MB transfer.

A/B built from the same tree, the only difference being the `record_download()` condition — 3 runs each:

| | before | after |
|---|---|---|
| download run 1 | 0.32 GB/s | 4.19 GB/s |
| download run 2 | 0.32 GB/s | 4.11 GB/s |
| download run 3 | 0.32 GB/s | 4.23 GB/s |

**0.32 → 4.19 GB/s, 13.1x.** For reference on the same machine: plain host `memcpy` 23.0 GB/s, ncnn upload 10.0 GB/s. The 31x upload/download asymmetry reported in the issue becomes ~2.4x.

## Tests

`ctest` (178 tests, `NCNN_VULKAN=ON`), run on two clean worktrees — pristine `a4d2ea1` and `a4d2ea1` + this commit:

| build | result |
|---|---|
| pristine `a4d2ea1` | 177/178 passed |
| this PR | 177/178 passed |

Identical failure sets. The single failure is `test_binaryop_3` → `test_binaryop_atan2_special op_type=10`, byte-identical output on both builds — a pre-existing atan2 special-value mismatch on RADV, unrelated to this change.

## One question

`VkCompute::record_clone(const VkMat& src, Mat& dst, const Option& opt)` (`command.cpp:649`) has the same shape of test:

```cpp
if (!src.allocator->mappable)
{
    // device to staging ...
}
```

and then defers a `TYPE_post_download` memcpy out of the mapped buffer — so on write-combined memory it should hit the identical uncached-read cost. I have **not** touched it here, since you asked specifically about the `record_download()` path and I have no benchmark isolating `record_clone`. Happy to add it to this PR or send it separately, whichever you prefer.

### DIFF
--- a/src/allocator.cpp
+++ b/src/allocator.cpp
@@ -355,6 +355,7 @@ VkAllocator::VkAllocator(const VulkanDevice* _vkdev)
     reserved_type_index = (uint32_t)-1;
     mappable = false;
     coherent = false;
+    cached = false;
 }
 
 VkAllocator::~VkAllocator()
@@ -779,6 +780,7 @@ VkBufferMemory* VkBlobAllocator::fastMalloc(size_t size)
 
         mappable = vkdev->is_mappable(buffer_memory_type_index);
         coherent = vkdev->is_coherent(buffer_memory_type_index);
+        cached = vkdev->is_cached(buffer_memory_type_index);
     }
 
     block->memory = allocate_memory(memoryRequirements.size, buffer_memory_type_index);
@@ -1058,6 +1060,7 @@ VkImageMemory* VkBlobAllocator::fastMalloc(int w, int h, int c, size_t elemsize,
 
         mappable = vkdev->is_mappable(image_memory_type_index);
         coherent = vkdev->is_coherent(image_memory_type_index);
+        cached = vkdev->is_cached(image_memory_type_index);
     }
 
     // create new block
@@ -1443,6 +1446,7 @@ VkBufferMemory* VkWeightAllocator::fastMalloc(size_t size)
 
                 mappable = vkdev->is_mappable(buffer_memory_type_index);
                 coherent = vkdev->is_coherent(buffer_memory_type_index);
+                cached = vkdev->is_cached(buffer_memory_type_index);
             }
 
             block->memory = allocate_dedicated_memory(memoryRequirements2.memoryRequirements.size, buffer_memory_type_index, 0, block->buffer);
@@ -1514,6 +1518,7 @@ VkBufferMemory* VkWeightAllocator::fastMalloc(size_t size)
 
                     mappable = vkdev->is_mappable(buffer_memory_type_index);
                     coherent = vkdev->is_coherent(buffer_memory_type_index);
+                    cached = vkdev->is_cached(buffer_memory_type_index);
                 }
 
                 block->memory = allocate_import_host_memory(memoryRequirements.size, buffer_memory_type_index, host_ptr);
@@ -1544,6 +1549,7 @@ VkBufferMemory* VkWeightAllocator::fastMalloc(size_t size)
 
                 mappable = vkdev->is_mappable(buffer_memory_type_index);
                 coherent = vkdev->is_coherent(buffer_memory_type_index);
+                cached = vkdev->is_cached(buffer_memory_type_index);
             }
 
             block->memory = allocate_memory(memoryRequirements.size, buffer_memory_type_index);
@@ -1596,6 +1602,7 @@ VkBufferMemory* VkWeightAllocator::fastMalloc(size_t size)
 
             mappable = vkdev->is_mappable(buffer_memory_type_index);
             coherent = vkdev->is_coherent(buffer_memory_type_index);
+            cached = vkdev->is_cached(buffer_memory_type_index);
         }
 
         block->memory = allocate_memory(memoryRequirements.size, buffer_memory_type_index);
@@ -1769,6 +1776,7 @@ VkImageMemory* VkWeightAllocator::fastMalloc(int w, int h, int c, size_t elemsiz
 
                 mappable = vkdev->is_mappable(image_memory_type_index);
                 coherent = vkdev->is_coherent(image_memory_type_index);
+                cached = vkdev->is_cached(image_memory_type_index);
             }
 
             // bind memory
@@ -1887,6 +1895,7 @@ VkImageMemory* VkWeightAllocator::fastMalloc(int w, int h, int c, size_t elemsiz
 
                     mappable = vkdev->is_mappable(image_memory_type_index);
                     coherent = vkdev->is_coherent(image_memory_type_index);
+                    cached = vkdev->is_cached(image_memory_type_index);
                 }
 
                 ptr->memory = allocate_import_host_memory(new_block_size, image_memory_type_index, host_ptr);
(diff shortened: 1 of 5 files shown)


tencent-adm:
[CLA assistant check](https://cla.opensource.tencent.com/Tencent/ncnn?pullRequest=6864) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.opensource.tencent.com/Tencent/ncnn?pullRequest=6864) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.opensource.tencent.com/check/Tencent/ncnn?pullRequest=6864) it.</sub>

PR #6865, opened 2026-07-27
vulkan: implement MatMul (fixes #6858)

### LINKED ISSUE #6858: Vulkan: MatMul has no Vulkan implementation — attention models fall back to CPU mid-graph (yolo12n 1.81x vs 5-8x for CNN peers)
### Summary

`MatMul` has no Vulkan implementation — there is no `src/layer/vulkan/matmul_vulkan.cpp`. Every `MatMul` in a Vulkan graph therefore runs on the CPU, which forces a download of its operands and a re-upload of its result in the middle of the graph.

This is the dominant remaining cost for attention-based models. `yolo12n` (ultralytics YOLO12, attention-centric) contains **16** `MatMul` layers and is by a wide margin the weakest model in a 22-model YOLO zoo on a Radeon 8060S:

| Model | raw forward pass speedup (Vulkan vs 4-thread CPU) | MatMul count |
|---|---|---|
| FastSAM-s | 8.70× | 0 |
| yolov8n | 5.40× | 0 |
| yolov8n-seg | 5.68× | 0 |
| yolo11n | 4.78× | 2 |
| **yolo12n** | **1.81×** | **16** |

Its per-layer GPU time is still 2.8× better than CPU (10.9 ms vs 30.4 ms measured with `-DNCNN_BENCHMARK=ON`), so the network itself is fine — the round trips consume the win.

Models affected in that zoo: every yolo11/yolo26 variant (2–4 MatMuls each) and yolo12n (16).

### Shapes that matter in practice

Attention blocks produce batched 3-D operands, e.g. from `yolo11n` at 640×640:

```
MatMul matmul_178     A (2,400,32)  B (2,32,400)   transB=0  -> (2,400,400)
MatMul matmultransb_0 A (2,64,400)  B (2,400,400)  transB=1  -> (2,64,400)
```

so a useful implementation needs the batched case, not just 2-D.

### I attempted this and did not succeed — notes in case they save someone time

I wrote a `MatMul_vulkan` delegating to a `matmul.comp` (one invocation per output element, broadcast batch strides as push constants). It is **correct for 2-D** (both `transB` values) but **wrong for batched shapes**, so I have not proposed it as a patch.

What I verified along the way, which may still be useful:

- **Dispatch and output addressing were correct.** A shader writing `b*10000 + m*100 + n` into the output reproduced the expected value for every element in every batch.
- **The dispatcher must be a real blob.** Passing a hand-built `VkMat` with only `w/h/c` set — `dims` and `elempack` left at their defaults — produced a z group count of 1, so every batch past the first was silently never written. Passing `top_blob` fixed that. This failure mode looks exactly like a stride bug, which cost me a while.
- **`support_packing = false`** does deliver `elempack=1` blobs as expected.

Where I got stuck is the per-channel stride of a 3-D `VkMat`. `concat.comp` uses `gz * psc(cstep) + gy * psc(w) + gx` with `bottom_blob.cstep`, so I used the same. An index probe reading `A[ab * cstep + m * K]` with a known input pattern showed that for an fp32 blob uploaded from a host `Mat`, `cstep` (36 for a 5×7 channel) lands **one element past** the true start of channel 1 (35) — the buffer appears tightly packed while `cstep` reports padding. For fp16 neither `cstep` nor `w*h` gave correct reads.

I suspect part of this is an artefact of how I built the test blobs (via the Python API from a numpy array, which wraps external memory and may produce a tightly-packed host `Mat` that `record_upload` preserves). But the real-graph path is also not right — `yolo11n` came out at corr 0.999998 while `yolo12n` was 0.998526, and disabling fp16 made `yolo12n` *worse* (0.9932), which rules out precision as the explanation.

So the question I could not answer is: **for a 3-D `VkMat` reaching a layer's `forward()`, what is the authoritative per-channel stride to use inside the shader, and does it differ between blobs produced by `record_upload` and blobs produced by a preceding layer?**

Happy to pick this back up with a pointer, or to test a patch.

### Environment

ncnn `a4d2ea1` (2026-07-22), Vulkan SDK 1.4.350.1, RADV, Ubuntu 24.04, Radeon 8060S Graphics (GFX1151).

giuliocorradi:
Implements the Vulkan `MatMul` requested in #6858.

`MatMul` had no Vulkan implementation — there is no `src/layer/vulkan/matmul_vulkan.cpp` — so every `MatMul` in a Vulkan graph ran on the CPU, forcing a download of both operands and a re-upload of the result mid-graph. On a 22-model YOLO zoo, `yolo12n` (16 `MatMul` layers) was the only model where **Vulkan was slower than the CPU**.

## Results

Raw forward pass, Radeon 8060S (Strix Halo, RADV GFX1151), 640×640, 4 threads, median of 3 alternating sessions, **same build on both sides**:

| model | MatMuls | Vulkan before | Vulkan after | |
|---|---|---|---|---|
| yolo12n | 16 | 62.64 ms | **8.93 ms** | **7.0x** |
| yolo11n | 2 | 9.77 ms | 6.02 ms | 1.6x |
| yolov8n | **0** | 5.99 ms | 5.93 ms | 1.01x |

`yolov8n` contains no `MatMul` and is included as a **control** — it does not move, which is what attributes the other two gains to this layer rather than to build or machine differences.

Two notes on honesty of measurement:
- These were taken **without** the #6857 download fix, so they isolate `MatMul` alone. With both applied the numbers improve further, but that would confound the two changes.
- The absolute figures are lower than the `1.81x -> 4.19x` I quoted in #6858; that earlier run was on a differently-loaded machine and a build that also carried the download fix. The before/after ratio here is measured in one sitting with alternating sessions.

## Correctness

`tests/test_matmul` passes. It runs the GPU path through `test_layer`, covering 1-D through 4-D operands, `transB`, and broadcasting.

Full `ctest`: **177/178 both with and without this commit**, identical failure sets. The single failure is `test_binaryop_3` → `test_binaryop_atan2_special op_type=10`, a pre-existing atan2 special-value mismatch on RADV, byte-identical on pristine `a4d2ea1`.

Worth flagging: my first draft **failed `test_matmul`**, and the failures were instructive enough to be worth recording here.

1. **Degenerate axes.** A 1-D operand contributes no dimension of its own. `MatMul::forward` computes into a `(N,1)` / `(1,M)` blob and reshapes the extra axis away. Creating `(N, M, …)` unconditionally gave `dims=2` where the CPU gives `dims=1`. The final shape is now created directly and described to the shader with explicit strides, because a reshaped blob does not carry the strides the general case implies.

2. **Rank promotion.** With mixed ranks, `MatMul::forward` lifts a 3-D operand with `reshape(w, h, c, 1)` — its **channel** axis becomes the **depth** axis, and depth then strides by the original `cstep`, not `w*h`. Reading `A.c` as channels rejected valid broadcasts: `test_matmul_13` (A `(14,23,10)` × B `(5,14,1,16)`) hit the broadcast guard, returned −1, and the never-created blob then segfaulted in `record_download`.

3. **Dispatch grid.** Dispatch is over the logical `(N, M, outd, outc)` grid rather than over the output blob, since a reshaped blob no longer carries `M` and `outd` separately while the shader decodes depth out of y as `gy / M`. The `Mat` dispatcher overload supplies the group counts without allocating.

## Implementation notes

- `support_vulkan_packing` stays at its default `false`: the shader indexes elements one at a time, which keeps batch slicing a simple stride instead of re-deriving pack4/pack8 addressing for every rank.
- Shape specialisation constants are left unspecialised, so shapes are resolved per call via push constants and one pipeline serves every shape.
- `ncnn_add_layer(MatMul)` already picks the Vulkan source up; no CMake change needed.

This is a straightforward scalar implementation — correct and a large improvement over the CPU round trip, but not a tuned GEMM. Tiling and cooperative matrix would go considerably further on the batched attention shapes from #6858. Happy to iterate if you would rather land something closer to that first.

### DIFF
--- a/src/layer/vulkan/matmul_vulkan.cpp
+++ b/src/layer/vulkan/matmul_vulkan.cpp
@@ -1,0 +1,247 @@
+// Copyright 2026 Tencent
+// SPDX-License-Identifier: BSD-3-Clause
+
+#include "matmul_vulkan.h"
+
+#include "layer_shader_type.h"
+
+namespace ncnn {
+
+MatMul_vulkan::MatMul_vulkan()
+{
+    support_vulkan = true;
+
+    // The shader indexes elements one at a time, so it needs the plain
+    // (unpacked) layout. support_vulkan_packing already defaults to false,
+    // which keeps batch slicing a simple stride and avoids re-deriving
+    // pack4/pack8 addressing for every rank.
+
+    pipeline_matmul = 0;
+}
+
+int MatMul_vulkan::create_pipeline(const Option& opt)
+{
+    std::vector<vk_specialization_type> specializations(1 + 3);
+    specializations[0].i = transB;
+    // Shapes are resolved per call, so leave the shape constants unspecialised.
+    specializations[1 + 0].i = 0;
+    specializations[1 + 1].i = 0;
+    specializations[1 + 2].i = 0;
+
+    pipeline_matmul = new Pipeline(vkdev);
+    pipeline_matmul->set_optimal_local_size_xyz(8, 8, 1);
+    pipeline_matmul->create(LayerShaderType::matmul, opt, specializations);
+
+    return 0;
+}
+
+int MatMul_vulkan::destroy_pipeline(const Option& /*opt*/)
+{
+    delete pipeline_matmul;
+    pipeline_matmul = 0;
+
+    return 0;
+}
+
+int MatMul_vulkan::forward(const std::vector<VkMat>& bottom_blobs, std::vector<VkMat>& top_blobs, VkCompute& cmd, const Option& opt) const
+{
+    const VkMat& A = bottom_blobs[0];
+    const VkMat& B = bottom_blobs[1];
+
+    const int Adims = A.dims;
+    const int Bdims = B.dims;
+
+    // Normalise both operands to (batch, rows, cols), mirroring MatMul::forward:
+    //   1-D A is a row vector (1, K); 1-D B is a column vector (K, 1).
+    const int M = Adims == 1 ? 1 : A.h;
+    const int K = A.w;
+    const int N = Bdims == 1 ? 1 : (transB == 0 ? B.w : B.h);
+
+    // c and d broadcast independently, exactly as MatMul::forward does.
+    //
+    // Mind the promotion: when the ranks are mixed, MatMul::forward lifts a 3-D
+    // operand with A.reshape(w, h, c, 1), which turns its CHANNEL axis into the
+    // DEPTH axis. Depth then strides by the original cstep rather than by w*h.
+    // Reading A.c as channels in that case rejects valid broadcasts.
+    const int max_ABdims = std::max(Adims, Bdims);
+
+    int a_c = 1, a_d = 1, a_cstep = 0, a_dstep = 0;
+    int b_c = 1, b_d = 1, b_cstep = 0, b_dstep = 0;
+
+    if (max_ABdims == 4 && Adims > 2)
+    {
+        if (Adims == 4)
+        {
+            a_c = A.c;
+            a_d = std::max(A.d, 1);
+            a_cstep = (int)A.cstep;
+            a_dstep = A.w * A.h;
+        }
+        else // Adims == 3, promoted to (w, h, c, 1)
+        {
+            a_d = A.c;
+            a_dstep = (int)A.cstep;
+        }
+    }
+    else if (Adims > 2)
+    {
+        a_c = A.c;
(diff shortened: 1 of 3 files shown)


tencent-adm:
[CLA assistant check](https://cla.opensource.tencent.com/Tencent/ncnn?pullRequest=6865) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.opensource.tencent.com/Tencent/ncnn?pullRequest=6865) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.opensource.tencent.com/check/Tencent/ncnn?pullRequest=6865) it.</sub>

PR #6866, opened 2026-07-27
pnnx: write bool attributes as fp32 so the ncnn .bin stays in step (fixes #6855)

### LINKED ISSUE #6855: pnnx: bool attributes desynchronise the ncnn .bin — MemoryData declares element count but only 1 byte/element is written
### Summary

When a model contains a **bool** attribute/constant, `pnnx` writes it to `model.ncnn.bin` as **one byte per element**, but emits a `MemoryData` layer that declares the element *count*. `MemoryData::load_model()` then reads that many **float32**. Because `ModelBin` is a sequential reader, the cursor ends up `3n/4` floats out of step and **every weight-bearing layer after that point loads shifted data**.

Nothing reports an error: `pnnx` exits 0, `load_param()` and `load_model()` both return 0, and inference produces plausible-looking numbers that are simply wrong.

### Environment

| | |
|---|---|
| pnnx | `20260526` (pip) |
| ncnn (python) | `1.0.20260526` |
| ncnn source | `a4d2ea1` (2026-07-22) |
| torch | 2.12.1 |
| OS | Ubuntu 24.04.3, x86_64 |
| Python | 3.10.20 |

### Minimal reproduction

```python
import subprocess, numpy as np, torch, torch.nn as nn, ncnn

N = 64   # 64 bytes written, 256 bytes read -> 48 floats adrift

class M(nn.Module):
    def __init__(self):
        super().__init__()
        self.c1 = nn.Conv2d(3, N, 3, padding=1)
        self.c2 = nn.Conv2d(N, 8, 3, padding=1)          # loaded AFTER the bool
        self.register_buffer("keep", torch.ones(1, N, 1, 1, dtype=torch.bool))

    def forward(self, x):
        return self.c2(self.c1(x) * self.keep)

torch.manual_seed(0)
m, x = M().eval(), torch.randn(1, 3, 32, 32)
ref = m(x).detach().numpy()[0]

np.save("in0.npy", x.numpy())
torch.jit.trace(m, x).save("m.pt")
subprocess.run(["pnnx", "m.pt", "input=in0.npy", "fp16=0",
                "ncnnparam=m.ncnn.param", "ncnnbin=m.ncnn.bin"])

net = ncnn.Net()
net.load_param("m.ncnn.param"); net.load_model("m.ncnn.bin")
ex = net.create_extractor()
ex.input("in0", ncnn.Mat(np.ascontiguousarray(x.numpy()[0])))
got = np.array(ex.extract("out0")[1])

print("corr", np.corrcoef(got.ravel(), ref.ravel())[0, 1])
```

Output:

```
pnnx exit status: 0            (no error reported)
ncnn load_param rc=0 load_model rc=0

torch out : shape=(8, 32, 32) range=[-1.6347, 1.2956]
ncnn  out : shape=(8, 32, 32) range=[-0.7004, 0.5721]
correlation: -0.037241   max|diff|=1.578086
```

The generated `.param` contains:

```
MemoryData   keep   0 1 <blob>  0=64
```

`64` elements declared → ncnn reads 256 bytes; pnnx wrote 64. The reader is 48 floats out of step from there on, so `c2`'s weights are garbage.

### Root cause

`tools/pnnx/src/pass_ncnn/convert_attribute.cpp` sets the `MemoryData` params from the tensor **shape**, without regard to dtype — correct for the layer, but it commits ncnn to reading `w*h*c` float32 (`MemoryData::load_model()` → `mb.load(..., load_type)`, default `load_type=1`, untagged fp32).

`tools/pnnx/src/save_ncnn.cpp` then writes the attribute. It special-cases fp32→fp16 and int64→int32:

```cpp
if (attr.type == 5) // i64 --> i32
{
    ...
    fwrite(data_int32.data(), data_int32.size() * sizeof(int32_t), 1, binfp);
    continue;
}

fwrite(attr.data.data(), attr.data.size(), 1, binfp);   // <-- bool lands here
```

There is no case for `attr.type == 9` (bool), so it takes the raw `fwrite` and emits 1 byte per element.

The mismatch is not unique to bool in principle — the raw path writes the tensor's native element size, while `MemoryData` always reads 4 bytes per element, so any narrower attribute dtype reaching this path has the same problem. bool is the one I hit and confirmed.

### Suggested fix

Convert bool (and other narrow dtypes) to float32 on write, mirroring the existing int64 case:

```cpp
if (attr.type == 9) // bool --> fp32
{
    const unsigned char* p = (const unsigned char*)attr.data.data();
    int len = attr.data.size();
    std::vector<float> data_fp32(len);
    for (int i = 0; i < len; i++)
        data_fp32[i] = p[i] ? 1.f : 0.f;
    fwrite(data_fp32.data(), data_fp32.size() * sizeof(float), 1, binfp);
    continue;
}
```

An alternative would be to keep the packed bytes and set `MemoryData`'s `load_type`, but there is no byte-wise load type, so widening on write looks like the smaller change.

### How this surfaced

Found while converting `yoloe-11s-seg-pf` (ultralytics YOLOE). A 400-element folded bool constant put the reader 300 floats out of step, which corrupted the anchor-point grid — it arrived as `[-0.032, -0.04, ...]` instead of `[0.5, 1.5, 2.5, ...]` — along with the mask and prototype convolutions. Output correlation against torch was 0.012. The class-score branch happened to sit *earlier* in the `.bin` and was unaffected, which made the corruption look selective and cost a lot of time to localise.

The silence is the expensive part: with a zero exit status and successful loads, there is no signal that the model is wrong until you compare against the reference implementation.

Happy to open a PR with the fix and a test case if that is useful.

giuliocorradi:
Fixes #6855.

A **bool** tensor attribute is written to `model.ncnn.bin` raw — one byte per element — but the `MemoryData` layer emitted for it is sized from the tensor **shape**, and `MemoryData::load_model()` reads that many **float32** (`load_type=1`, untagged).

`ModelBin` is a sequential reader, so after an n-element bool the cursor is `3n/4` floats out of step and **every weight-bearing layer after that point loads shifted data**. Nothing reports it: `pnnx` exits 0, `load_param()` and `load_model()` both return 0, and inference produces plausible numbers that are simply wrong.

The fix widens bool to fp32 on write, mirroring the `i64 -> i32` handling directly above it.

## Correction to the repro in the issue

**The reproduction I posted in #6855 does not actually trigger the bug** — I should have checked it before filing. In that snippet the bool is a `register_buffer` consumed by a multiply, so **torch type-promotes it to float during tracing** and a type-9 attribute never reaches `save_ncnn`. Both patched and unpatched builds emit an identical 13224-byte `.bin` for it.

The real trigger is a bool that is *folded into a constant* and stays bool. A minimal one:

```python
class M(nn.Module):
    def __init__(self):
        super().__init__()
        self.c1 = nn.Conv2d(3, 400, 1)
        self.c2 = nn.Conv2d(400, 8, 1)
        self.register_buffer("t", torch.randn(1, 400, 1, 1))
    def forward(self, x):
        m = self.t > 0                      # folds to a bool CONSTANT
        y = self.c1(x)
        return self.c2(torch.where(m, y, torch.zeros_like(y)))
```

`torch.logical_and(self.a > 0, self.b > 0)` in place of the comparison behaves identically.

## Evidence

Two `pnnx` binaries built from **identical source**, differing only in `save_ncnn.cpp`:

| construct | unpatched | patched | delta |
|---|---|---|---|
| `torch.where` + folded bool | 113240 | 114440 | **+1200** |
| `logical_and` + folded bool | 113240 | 114440 | **+1200** |

400 elements × (4 − 1) bytes = 1200, exactly the shortfall.

On a real model — an ultralytics YOLOE export, which is where I hit this — the graph carries `MemoryData pnnx_fold_idx.1 0 1 335 0=400`, a 400-element folded bool, and the same +1200 delta appears (47467496 → 47468696). There the shifted reads corrupted the anchor grid, which arrived as `[-0.032, -0.04, ...]` instead of `[0.5, 1.5, 2.5, ...]`, along with the mask and proto convolutions. Layers *earlier* in the file were untouched, which is what makes the damage look selective and misleading.

One caveat on method, since it nearly misled me: I first tried to A/B the PyPI wheel against a locally built binary. That is **not** a valid comparison — the two differ in more than this patch (they emit different blob numbering and a different `pnnx_188` shape), so the byte-level tail comparison does not line up. The numbers above are from same-source builds only.

## Note for a runnable check

The minimal repro above exercises the writer but produces a graph containing `torch.where`, which ncnn has no layer for, so it cannot be executed end to end to show wrong *output* — only the wrong `.bin` size. The end-to-end corruption is visible on the YOLOE export.

### DIFF
--- a/tools/pnnx/src/save_ncnn.cpp
+++ b/tools/pnnx/src/save_ncnn.cpp
@@ -288,6 +288,26 @@ int save_ncnn(const Graph& g, const std::string& parampath, const std::string& b
                 continue;
             }
 
+            if (attr.type == 9) // bool --> fp32
+            {
+                // MemoryData is emitted from the tensor SHAPE and ncnn reads
+                // w*h*c float32 (load_type=1, untagged). Writing bools raw is one
+                // byte per element, so the sequential ModelBin reader ends up 3n/4
+                // floats out of step and EVERY later weight-bearing layer loads
+                // shifted data -- silently, with a zero exit status. Widen on write
+                // so what is written matches what will be read.
+                const unsigned char* p = (const unsigned char*)attr.data.data();
+                const size_t len = attr.data.size();
+                std::vector<float> data_fp32(len);
+                for (size_t i = 0; i < len; i++)
+                {
+                    data_fp32[i] = p[i] ? 1.f : 0.f;
+                }
+
+                fwrite(data_fp32.data(), data_fp32.size() * sizeof(float), 1, binfp);
+                continue;
+            }
+
             fwrite(attr.data.data(), attr.data.size(), 1, binfp);
         }
 


tencent-adm:
[CLA assistant check](https://cla.opensource.tencent.com/Tencent/ncnn?pullRequest=6866) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.opensource.tencent.com/Tencent/ncnn?pullRequest=6866) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.opensource.tencent.com/check/Tencent/ncnn?pullRequest=6866) it.</sub>

nihui:
superseded by https://github.com/Tencent/ncnn/pull/7006

PR #6869, opened 2026-07-27
fix MHA singleton mask broadcast across backends

futz12:
## Summary

  This PR fixes persistent KV-cache allocation and singleton 3D attention-mask broadcasting across the generic, x86, ARM, MIPS,
  LoongArch, and Vulkan attention implementations.

  It also fixes the equivalent singleton-mask issue in Vulkan SDPA.

  ## Problems

  ### Singleton 3D masks were treated as per-head masks

  An attention mask with shape:

  (dst_seqlen, src_seqlen, 1)

  is intended to be broadcast across every attention head.

  Several implementations treated every 3D mask as a per-head mask and selected:

  attn_mask_blob.channel(head)

  For head > 0, this accesses a channel that does not exist.

  Affected paths included:

  - generic MHA FP32
  - generic MHA INT8
  - generic weight block-quantized MHA
  - x86 MHA
  - ARM MHA
  - MIPS MHA
  - LoongArch MHA
  - Vulkan MHA
  - Vulkan SDPA normal path
  - Vulkan SDPA Flash Attention path
  - Vulkan SDPA cooperative-matrix path

  Singleton masks now always select channel zero, while masks with multiple channels retain their per-head behavior.

  ### Vulkan mask indexing

  The Vulkan SDPA shaders use the following addressing logic:

  mask_head = attn_mask_dims == 3 ? head : 0;
  mask_index = mask_head * mask_cstep + row * dst_seqlen + column;

  Previously, a dims=3, c=1 mask was still passed to the shader as a 3D per-head mask. Every head greater than zero therefore
  read beyond the only allocated mask channel.

  This was reproduced on an NVIDIA RTX 4060:

  past_seqlen=2560
  num_heads=4
  mask_channels=1

  value not match at c:1 h:0 w:1
  expect 0.016805 but got 0.015332

  The old implementation failed with exit code 1. After treating a singleton 3D mask as a 2D broadcast mask, the same test passes
  with exit code 0.

  Vulkan MHA additionally keeps 3D masks unpacked so that head packing cannot be confused with query-position packing.

  ## Regression coverage

  The MHA and SDPA tests now cover:

  - single-token decoding
  - multi-token decoding with 3 and 4 tokens
  - long KV caches up to 2560 tokens
  - singleton masks
  - per-head masks
  - INT8 MHA
  - weight block-quantized MHA
  - cross-attention cache reuse
  - 257-step rolling KV-cache decoding
  ## Verification

  The following tests pass:

  test_multiheadattention_kvcache       CPU: PASS
  test_sdpa_kvcache                     CPU: PASS
  test_multiheadattention_block_quant   CPU: PASS
  test_multiheadattention_kvcache       Vulkan: PASS
  test_sdpa_kvcache                     Vulkan: PASS


lazypool (commented):
Review compeleted: This addresses two distinct edge cases: persistent KV-cache tensors now use the caller-provided blob allocator, and 3D masks with a single channel are treated as shared across heads. The added CPU and Vulkan regression coverage, including long-cache cases, helps validate both behaviors.

PR #6882, opened 2026-08-01
fix spirv compile failure diagnostics read past shader buffer

### LINKED ISSUE #6881: Out-of-bounds read in compile_spirv_module()'s error-printing path - on non-null-terminated shader data
## Description

When a Vulkan compute shader fails to compile via compile_spirv_module()
(src/gpu.cpp), the diagnostic error-printing code treats the shader source as
a null-terminated C string, but it isn't one — this causes an out-of-bounds
read that produces garbled, misleading diagnostic output, and reads past the
end of the intended buffer into unrelated adjacent memory.

## Root cause

In src/gpu.cpp, compile_spirv_module()'s failure path (currently ~line
6337-6353 on master):

```
const char* p = comp_datas[3];
const char* line_end;
int line_number = 1;

while ((line_end = strchr(p, '\n')) != NULL)
{
    NCNN_LOGE("%d:\t%.*s", line_number++, (int)(line_end - p), p);
    p = line_end + 1;
}

if (*p != '\0')
{
    NCNN_LOGE("%d:\t%s", line_number, p);
}
```
comp_datas[3] points into a static const char whatever_comp_data[] array,
generated at build time by cmake/ncnn_generate_shader_comp_header.cmake. That
script hex-encodes the raw shader source bytes with no terminating NUL byte 
added:

```
file(WRITE ${SHADER_COMP_HEADER} "static const char 
${SHADER_SRC_NAME_WE}_comp_data[] = {${comp_data_hex}};\n")
```

Since strchr() and the *p != '\0' check both require a null-terminated string,
and none is guaranteed here, both calls can read past the end of the intended
array. In practice these per-shader arrays are laid out contiguously in
.rodata by the linker, so strchr often walks harmlessly through the start of
the next embedded shader's data (which conveniently starts with #version,
ASCII #) — but it isn't bounded, and can end up printing arbitrary adjacent
memory.

## Impact

It corrupts the diagnostic output every time any shader fails to compile, on any platform - making it
harder to debug the real underlying problem.
In principle it's also a genuine out-of-bounds read, not just a cosmetic issue.

## Suggested fix

The exact length of comp_datas[3] is already known (comp_data_sizes[3],
computed earlier in the same function) — use it instead of relying on
null-termination, e.g. bound the scan with memchr()/an explicit end pointer
rather than strchr(), or simply add a terminating NUL byte when generating the
.comp_data[] arrays.

## Environment

Reproduced with ncnn release 20260526, NCNN_SYSTEM_GLSLANG=ON, but the buggy
code is unchanged on current master and the bug isn't specific to that build
flag — it can be triggered by any shader compile failure on any platform,
since it's purely in the error-printing path.

lazypool:
### Summary

Fixes #6881 — the error-printing path in `compile_spirv_module()` (`src/gpu.cpp`) treated the shader source slice as a null-terminated C string when it is not, causing an out-of-bounds read and garbled diagnostic output whenever a Vulkan compute shader fails to compile.

### Root cause

- Shader `.comp_data[]` arrays are generated by `cmake/ncnn_generate_shader_comp_header.cmake` as hex-encoded bytes **without a trailing `0x00`**.
- `compile_spirv_module()` slices the source into `comp_datas[3]` (everything after `#version`) and, on parse failure, scans it with `strchr()` / `*p != '\0'` — both require null-termination, but none is guaranteed.
- In practice `strchr()` walks past the end of the buffer into adjacent memory, printing garbage (or, with ASan, reporting a `stack-buffer-overflow`).

### Change

Bound the scan using the already-known source length (`comp_data_size_2`, the size of `comp_datas[3]`):

- `strchr(p, '\n')` → `memchr(p, '\n', p_end - p)`
- final `if (*p != '\0')` → `if (p != p_end)`
- final `%s` print → `%.*s` with the remaining length

### Verification

Built with `-DNCNN_VULKAN=ON` (system glslang 16.3.0) and tested under AddressSanitizer on RADV:

- **Before**: compiling a broken, non-null-terminated shader source via `compile_spirv_module(data, size, ...)` → `ERROR: AddressSanitizer: stack-buffer-overflow ... in ncnn::compile_spirv_module(...)`
- **After**: same input prints the shader lines cleanly and bounded (`1:\tvoid main() { ... }`), no ASan report; a real shader still compiles (`rc=0`).

No behavioral change to successful compilation paths.

### DIFF
--- a/src/gpu.cpp
+++ b/src/gpu.cpp
@@ -6340,18 +6340,19 @@ int compile_spirv_module(const char* comp_data, int comp_data_size, const Option
             // print as line_number: code
             {
                 const char* p = comp_datas[3];
+                const char* p_end = p + comp_data_size_2;
                 const char* line_end;
                 int line_number = 1;
 
-                while ((line_end = strchr(p, '\n')) != NULL)
+                while ((line_end = (const char*)memchr(p, '\n', p_end - p)) != NULL)
                 {
                     NCNN_LOGE("%d:\t%.*s", line_number++, (int)(line_end - p), p);
                     p = line_end + 1;
                 }
 
-                if (*p != '\0')
+                if (p != p_end)
                 {
-                    NCNN_LOGE("%d:\t%s", line_number, p);
+                    NCNN_LOGE("%d:\t%.*s", line_number, (int)(p_end - p), p);
                 }
             }
 


tencent-adm:
[CLA assistant check](https://cla.opensource.tencent.com/Tencent/ncnn?pullRequest=6882) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you all sign our [Contributor License Agreement](https://cla.opensource.tencent.com/Tencent/ncnn?pullRequest=6882) before we can accept your contribution.<br/>**0** out of **2** committers have signed the CLA.<br/><br/>:x: lazypool<br/>:x: nihui<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.opensource.tencent.com/check/Tencent/ncnn?pullRequest=6882) it.</sub>

lazypool (commented):
Please review when you get a chance, thanks!

nihui (approved):

nihui:

lazypool:
@nihui Thanks for your review. :-D

PR #6899, opened 2026-08-11
fix bf16p shader compile failure

### LINKED ISSUE #6883: Vulkan bf16_packed shader compile failure on shaderInt16-less GPUs
bf16_packed selects uint16_t without checking shaderInt16, causing Vulkan shader compile failure on GPUs that support fp16 storage/uniform but not shaderInt16 (e.g. Raspberry Pi 4/5, Mesa v3dv)

## error log

```
compile spir-v module failed
ERROR: 3:7: '' :  syntax error, unexpected IDENTIFIER
ERROR: 1 compilation errors.  No code generated.
compile_spirv_module failed -1
create_shader_module failed
binding_count not match, expect 0 but got 4 + 0
push_constant_count not match, expect 48 but got 3
CMake Error at .../cmake/run_test.cmake:4 (message):
  Test failed with return value 'SIGTRAP'
```

(`push_constant_count`/`binding_count` values vary run to run - they come from reading a `ShaderInfo` struct left uninitialized after the failed compile, not from anything meaningful. The real error is the `syntax error, unexpected IDENTIFIER` above it.)

## context

- ncnn, current `master` (commit `a4d2ea1`)
- Build flags: `-DNCNN_VULKAN=ON -DNCNN_SYSTEM_GLSLANG=ON` (shaders compiled from GLSL text at runtime via system `glslang`, not precompiled SPIR-V)
- Hardware: Raspberry Pi 4B, Broadcom VideoCore VI GPU
- Driver: Mesa `v3dv` (Mesa 25.1.8), confirmed via `vulkaninfo`: `driverID = DRIVER_ID_MESA_V3DV`
- This is not OS specific: `shaderInt16 = false` is a hardcoded literal in Mesa's own `v3dv_device.c` (`VkPhysicalDeviceFeatures` initializer)
- Confirmed via `ncnn::get_gpu_info()`: this device reports `support_fp16_storage=1`, `support_bf16_packed=1`, `physicalDevicefeatures().shaderInt16=0`.

## how to reproduce
1. Build ncnn with `NCNN_VULKAN=ON` on a device where `shaderInt16=false` but `support_fp16_storage=true` (Raspberry Pi 4/5 with Mesa `v3dv` is a ready-made example; any GPU with the same capability combination will do).
2. Run any layer test whose option matrix includes the row `{pack=1, fp16p=0, fp16s=0, fp16a=0, bf16p=1, bf16s=0}` - present unconditionally in `tests/testutil.cpp`'s `test_layer()` matrix, no device-capability gating. For example: `ctest -R test_absval`.
3. Observe the compile failure above. It is 100% deterministic - reproduces on a single isolated `ctest -R` run, with `OMP_NUM_THREADS=1`, at `-O0`, and with the vendored `glslang` submodule instead of system `glslang` (all tested to rule out unrelated causes).

## more

**Root cause**: in `compile_spirv_module()` (`src/gpu.cpp`), four separate `else if (opt.use_bf16_packed)` branches each independently select a `uint16_t`/`u16vec2`-based macro variant, guarded only by `support_fp16_storage` or `support_fp16_uniform`:

```cpp
else if (opt.use_bf16_packed)
{
    if (support_fp16_storage)
    {
        custom_defines.append("sfp", "uint16_t");
    }
    else
    {
        custom_defines.append("sfp", "uint");
    }
    ...
```

(the other three: the `lfp` type; the `lfp2afp`/`afp2lfp` conversion macros, where `afp2lfp` explicitly constructs `uint16_t(...)`; and the `buffer_ld1`/`buffer_st1`/`buffer_cp1`/`buffer_cp1to4`/`buffer_cp4to1` helper macros, which also use `u16vec2`/`pack32()`)

None of these four checks consult the device's actual 16-bit-integer-arithmetic support (available in the same function as `support_int16_arithmetic`/`support_shader_int16` = `info.physicalDevicefeatures().shaderInt16`, and already used correctly elsewhere in the function to gate the `GL_EXT_shader_explicit_arithmetic_types_int16` extension pragma - just not consulted when *selecting* the type in the first place). The existing `uint`/`float` fallback in each branch is already correct for exactly this situation; it's just never reached when it should be.

Confirmed via a Vulkan-free standalone `glslang::TShader::parse()` reproducer (no `VkInstance`/`VkDevice` ever created, linking only `libglslang`/`libSPIRV*`) that this is purely a shader-compiler-level issue, not a driver/loader problem - the failure reproduces identically outside of any Vulkan context.

**Suggested fix**

in each of the four branches:

```cpp
if (support_fp16_storage)   ->  if (support_fp16_storage && support_int16_arithmetic)
if (support_fp16_uniform)   ->  if (support_fp16_uniform && support_int16_arithmetic)
```

[001-fix-bf16-packed-missing-int16-extension-check-master.patch](https://github.com/user-attachments/files/30619048/001-fix-bf16-packed-missing-int16-extension-check-master.patch)

aadpnde:
Based on @marcin-sochacki idea for a fix for #6883. Since #6895 was
merged in, the checks need to be made to ncnn_glsl_ext.comp instead of
gpu.cpp.

Also added 2 buffer checks based on @marcin-sochacki's report since
there were 2 buffer_sm1 that were still reading the buffer as if it
had a 16bit value in each entry. And updated gpu.cpp so the 16 bit
shaders ext are only enabled when shaderInt16 is supported

Fixes #6883

### DIFF
--- a/src/gpu.cpp
+++ b/src/gpu.cpp
@@ -5800,7 +5800,7 @@ int compile_spirv_module(const char* comp_data, int comp_data_size, const Option
     {
         custom_exts += "#extension GL_EXT_shader_explicit_arithmetic_types_int64: require\n";
     }
-    if (support_shader_int16 || (opt.use_bf16_packed && support_fp16_storage))
+    if (support_shader_int16)
     {
         custom_exts += "#extension GL_EXT_shader_explicit_arithmetic_types_int16: require\n";
     }
@@ -5808,7 +5808,7 @@ int compile_spirv_module(const char* comp_data, int comp_data_size, const Option
     {
         custom_exts += "#extension GL_EXT_bfloat16: require\n";
     }
-    if (opt.use_fp16_storage || opt.use_bf16_storage || opt.use_int16_storage || (opt.use_bf16_packed && support_fp16_storage))
+    if (opt.use_fp16_storage || opt.use_bf16_storage || opt.use_int16_storage || (opt.use_bf16_packed && support_fp16_storage && support_shader_int16))
     {
         custom_exts += "#extension GL_EXT_shader_16bit_storage: require\n";
     }
--- a/src/ncnn_glsl_ext.comp
+++ b/src/ncnn_glsl_ext.comp
@@ -9,7 +9,7 @@
 #define unpackBFloat2x16(v) vec2(uintBitsToBFloat16EXT(unpackUint2x16(v)))
 #define packBFloat2x16(v) packUint2x16(bfloat16BitsToUintEXT(bf16vec2(v)))
 #elif NCNN_bf16_packed
-#if ncnn_storageBuffer16BitAccess
+#if ncnn_storageBuffer16BitAccess && ncnn_shaderInt16
 #define sfp uint16_t
 #else
 #define sfp uint
@@ -59,7 +59,7 @@
 #define lfp bfloat16_t
 #define lfpvec4 bf16vec4
 #elif NCNN_bf16_packed
-#if ncnn_uniformAndStorageBuffer16BitAccess
+#if ncnn_uniformAndStorageBuffer16BitAccess && ncnn_shaderInt16
 #define lfp uint16_t
 #else
 #define lfp float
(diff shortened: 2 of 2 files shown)


tencent-adm:
[CLA assistant check](https://cla.opensource.tencent.com/Tencent/ncnn?pullRequest=6899) <br/>All committers have signed the CLA.

nihui:
superseded by https://github.com/Tencent/ncnn/pull/6962

PR #6902, opened 2026-08-15
fix out-of-bounds blob index and unvalidated count in load_param_bin

kaltsit-10:
## Summary

`ncnn::Net::load_param_bin()` reads attacker-controlled integers from a `.parambin`
model file and uses them (a) to index into `d->blobs` with no bounds check, and
(b) to size `layer->bottoms` / `layer->tops` with no sign check. A crafted file can:

1. set `bottom_blob_index` / `top_blob_index` to any int32 → out-of-bounds read
   and write (`Blob& blob = d->blobs[idx]; blob.consumer = i;`) — CWE-125/787
2. set `bottom_count` / `top_count` to a negative value → `std::vector::resize(-1)`
   throws `std::length_error` → uncaught → `std::terminate()` — DoS

## Changes

- validate `bottom_count` / `top_count` are non-negative before use
- validate `bottom_blob_index` / `top_blob_index` are in `[0, blob_count)` before
  indexing `d->blobs`

## Why

ncnn parses untrusted model files in many embedding scenarios (mobile/desktop apps,
model hubs, WeChat/QQ). On master, a 149-byte crafted `.parambin` drives a
deterministic SIGSEGV (ASan: WRITE @ `src/net.cpp:1810`, `blob.consumer = i`);
a 28-byte file with `top_count = -1` aborts via uncaught `std::length_error`.
Both are fully reproducible and independent of the already-fixed CVE-2026-50144,
whose fix (`5a0288f2`) only guards parameter ids in `src/paramdict.cpp`.

## Test

| input | before (master a4d2ea1) | after |
|---|---|---|
| crafted `bottom_blob_index=0x09090909` (149B) | SIGSEGV (exit 139) | `ret=-1`, no crash |
| crafted `top_count=-1` (28B) | abort (exit 134, `std::length_error`) | `ret=-1`, no crash |
| valid header file | `ret=-1` (ParamDict err) | identical, no false positive |

The `clear(); return -1;` error path matches existing failure handling in this
function.
```

### DIFF
--- a/src/net.cpp
+++ b/src/net.cpp
@@ -24,6 +24,12 @@
 
 namespace ncnn {
 
+// maximum sizes accepted from an untrusted model file header;
+// larger values indicate a corrupt or malicious model and are rejected
+// before any allocation (resize with a huge count would otherwise OOM)
+static const int MAX_LAYER_COUNT = 1000000;
+static const int MAX_BLOB_COUNT = 1000000;
+
 class NetPrivate
 {
 public:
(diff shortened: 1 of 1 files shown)


tencent-adm:
[CLA assistant check](https://cla.opensource.tencent.com/Tencent/ncnn?pullRequest=6902) <br/>All committers have signed the CLA.

kaltsit-10:
Binary-path guards are now merged into #6922

nihui:
superseded by https://github.com/Tencent/ncnn/pull/6989

PR #6905, opened 2026-08-16
fix(allocator): drop _POSIX_C_SOURCE from posix_memalign branch

### LINKED ISSUE #6888: src/allocator.h's fastMalloc()/fastFree() pick between two mutually incompatible allocation schemes based on the preprocessor check _POSIX_C_SOURCE >= 200112L
## Context
```
#elif (defined(__unix__) || defined(__APPLE__)) && _POSIX_C_SOURCE >= 200112L ...
    // scheme A: posix_memalign() / plain free(ptr)
#else
    // scheme B: malloc(size + extra) + manual pointer-offset alignment,
    //           storing a hidden "real pointer" 8 bytes before the
    //           returned pointer; fastFree reads it back before free()
```

Mat::release() (which calls fastFree()) is NCNN_FORCEINLINE — defined in the header, so it is recompiled fresh in every translation unit that includes mat.h, unlike Mat::create()/fastMalloc()'s caller path, which in practice is always reached through mat.cpp's single compiled copy inside libncnn.so.

mat.cpp (compiled without ever including Python.h) leaves _POSIX_C_SOURCE undefined→ scheme B.
python/src/main.cpp (any pybind11 binding TU, including ncnn's own) includes <pybind11/pybind11.h> → <Python.h> → pyconfig.h, which unconditionally #defines _POSIX_C_SOURCE 200809L before mat.h is ever reached → scheme A.

So Mat::create() allocates via scheme B (storing a hidden back-pointer), but the inlined Mat::release() compiled into the Python module frees via scheme A's plain free(ptr) - passing the aligned data pointer itself to free(), which was never a value any allocator function actually returned. free() on that bogus pointer deterministically corrupts the heap and aborts. This is unconditional and 100% reproducible, on every architecture where mat.cpp's default _POSIX_C_SOURCE state differs from what Python.h forces.

## The fix

Drop the _POSIX_C_SOURCE condition entirely - posix_memalign/free are available on any real POSIX system regardless of that macro, and the guard serves no purpose other than making the branch selection dependent on include order:
```
-#elif (defined(__unix__) || defined(__APPLE__)) && _POSIX_C_SOURCE >= 200112L || (__ANDROID__ && __ANDROID_API__ >= 17)
+#elif (defined(__unix__) || defined(__APPLE__)) || (__ANDROID__ && __ANDROID_API__ >= 17)
```

ktz03:
Makes `fastMalloc`/`fastFree` use a consistent posix_memalign/free scheme regardless of whether `_POSIX_C_SOURCE` was defined by an earlier include (e.g. Python.h).

Fixes #6888


tencent-adm:
[CLA assistant check](https://cla.opensource.tencent.com/Tencent/ncnn?pullRequest=6905) <br/>All committers have signed the CLA.

Issue #6915, opened 2026-08-17
bug: heap-buffer-overflow (read) in ParamDict::load_param string parsing (src/paramdict.cpp:387)


beilzx:
## error log

```
find_blob_index_by_name fire2/squeejsqueeze1x1 failed
find_blob_index_by_name fire4/squeeze1x1_fire4/delu_squeeze1x1 failed
find_blob_index_by_name 1 failed
=================================================================
==3944598==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x50300000021f at pc 0x556938c2967d bp 0x7ffe1a5514e0 sp 0x7ffe1a5514d8
READ of size 1 at 0x50300000021f thread T0
    #0 0x556938c2967c in ncnn::ParamDict::load_param(ncnn::DataReader const&) /src/test/results/fuzzing_ncnn/test/ncnn/src/paramdict.cpp:387:61
    #1 0x556938bf0302 in ncnn::Net::load_param(ncnn::DataReader const&) /src/test/results/fuzzing_ncnn/test/ncnn/src/net.cpp:1477:33
    #2 0x556938c19fee in ncnn::Net::load_param(_IO_FILE*) /src/test/results/fuzzing_ncnn/test/ncnn/src/net.cpp:2177:22
    #3 0x556938bcdbdb in main /src/test/fuzzing_ncnn/harness.cpp:17:9
    #4 0x7fb86460e1c9  (/lib/x86_64-linux-gnu/libc.so.6+0x2a1c9) (BuildId: 274eec488d230825a136fa9c4d85370fed7a0a5e)
    #5 0x7fb86460e28a in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x2a28a) (BuildId: 274eec488d230825a136fa9c4d85370fed7a0a5e)
    #6 0x556938aea194 in _start (/src/test/results/fuzzing_ncnn/test/ncnn/build/ncnn_param+0x13e194) (BuildId: 8489885d434959364ed3a3e9d6ed984798b5dc8d)

0x50300000021f is located 1 bytes before 31-byte region [0x503000000220,0x50300000023f)
allocated by thread T0 here:
    #0 0x556938bcb641 in operator new(unsigned long) (/src/test/results/fuzzing_ncnn/test/ncnn/build/ncnn_param+0x21f641) (BuildId: 8489885d434959364ed3a3e9d6ed984798b5dc8d)
[...]

Issue #6914, opened 2026-08-17
bug: heap-buffer-overflow (write) in Mat::create via ParamDict::load_param (src/mat.cpp:337)


beilzx:
## error log

```
find_blob_index_by_name 56 failed
find_blob_index_by_name 57 failed
find_blob_index_by_name 58 failed
find_blob_index_by_name 59 failed
find_blob_index_by_name 60 failed
find_blob_index_by_name 61 failed
find_blob_index_by_name fiat failed
find_blob_index_by_name fire3/s,3,74,74,641 failed
find_blob_index_by_name 66 failed
find_blob_index_by_name 8=2 failed
find_blob_index_by_name 9=1 failed
find_blob_index_by_name Co failed
find_blob_index_by_name co failed
find_blob_index_by_name 3/relu_expand1x1 failed
find_blob_index_by_name firelu_expand3x3 failed
find_blob_index_by_name fire3/concat failed
find_blob_index_by_name -0=4,3,74,74,128 failed
find_blob_index_by_name Pooling failed
find_blob_index_by_name ol failed
find_blob_index_by_name pool3 failed
find_blob_index_by_name -23330=4,3,37,37,128 failed
find_blob_index_by_name Convolution failed
find_blob_index_by_name fire4/squeeze1x1 failed
find_blob_index_by_name fire4/squeeze1x1_fire4/relu_squeeze1x1 failed
find_blob_index_by_name -23330=4,3,37,37,32 failed
find_blob_index_by_name 0=32 failed
find_blob_index_by_name 1=4096 failed
find_blob_index_by_name 8=102 failed
find_blob_index_by_name Split failed
find_blob_index_by_name _3 failed
find_blob_index_by_name fire4/squlitncnn_0 failed
find_blob_index_by_name f1 failed
find_blob_index_by_name -2333037,32,3,37,37,32 failed
find_blob_index_by_name x1 failed
find_blob_index_by_name fire4/squeeze1nn_1 failed
find_blob_index_by_name firxpanexpand1x1_ failed
find_blob_index_by_name 9=firelu_expand3x3 failed
=================================================================
==3172928==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x509000000020 at pc 0x5602b9485c8d bp 0x7ffec93884f0 sp 0x7ffec93884e8
WRITE of size 4 at 0x509000000020 thread T0
[...]

Issue #6913, opened 2026-08-17
bug: heap-buffer-overflow (read/write) in Net::load_param blob index out-of-bounds (src/net.cpp:1464)


beilzx:
## error log

```
find_blob_index_by_name Input failed
=================================================================
==3244058==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x511000000280 at pc 0x55e7afa62989 bp 0x7ffdc29e35d0 sp 0x7ffdc29e35c8
READ of size 8 at 0x511000000280 thread T0
    #0 0x55e7afa62988 in std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char>>::_M_data() const /usr/include/c++/13/bits/basic_string.h:223:28
    #1 0x55e7afa62988 in std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char>>::_M_is_local() const /usr/include/c++/13/bits/basic_string.h:264:13
    #2 0x55e7afa62988 in std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char>>::operator=(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char>>&&) /usr/include/c++/13/bits/basic_string.h:858:18
    #3 0x55e7afa62988 in ncnn::Net::load_param(ncnn::DataReader const&) /src/test/results/fuzzing_ncnn/test/ncnn/src/net.cpp:1464:46
    #4 0x55e7afa7bfee in ncnn::Net::load_param(_IO_FILE*) /src/test/results/fuzzing_ncnn/test/ncnn/src/net.cpp:2177:22
    #5 0x55e7afa2fbdb in main /src/test/fuzzing_ncnn/harness.cpp:17:9
    #6 0x7fe2811771c9  (/lib/x86_64-linux-gnu/libc.so.6+0x2a1c9) (BuildId: 274eec488d230825a136fa9c4d85370fed7a0a5e)
    #7 0x7fe28117728a in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x2a28a) (BuildId: 274eec488d230825a136fa9c4d85370fed7a0a5e)
    #8 0x55e7af94c194 in _start (/src/test/results/fuzzing_ncnn/test/ncnn/build/ncnn_param+0x13e194) (BuildId: 8489885d434959364ed3a3e9d6ed984798b5dc8d)

[...]

Issue #6911, opened 2026-08-17
bug: FPE (divide-by-zero) in ConvolutionDepthWise::load_param (src/layer/convolutiondepthwise.cpp:46)


beilzx:
## error log

```
AddressSanitizer:DEADLYSIGNAL
=================================================================
==3215629==ERROR: AddressSanitizer: FPE on unknown address 0x5648463adbbb (pc 0x5648463adbbb bp 0x7ffd3666aa50 sp 0x7ffd3666a890 T0)
    #0 0x5648463adbbb in ncnn::ConvolutionDepthWise::load_param(ncnn::ParamDict const&) /src/test/results/fuzzing_ncnn/test/ncnn/src/layer/convolutiondepthwise.cpp:46:20
    #1 0x564844c40abc in ncnn::Net::load_param(ncnn::DataReader const&) /src/test/results/fuzzing_ncnn/test/ncnn/src/net.cpp:1524:35
    #2 0x564844c69fee in ncnn::Net::load_param(_IO_FILE*) /src/test/results/fuzzing_ncnn/test/ncnn/src/net.cpp:2177:22
    #3 0x564844c1dbdb in main /src/test/fuzzing_ncnn/harness.cpp:17:9
    #4 0x7ff2814461c9  (/lib/x86_64-linux-gnu/libc.so.6+0x2a1c9) (BuildId: 274eec488d230825a136fa9c4d85370fed7a0a5e)
    #5 0x7ff28144628a in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x2a28a) (BuildId: 274eec488d230825a136fa9c4d85370fed7a0a5e)
    #6 0x564844b3a194 in _start (/src/test/results/fuzzing_ncnn/test/ncnn/build/ncnn_param+0x13e194) (BuildId: 8489885d434959364ed3a3e9d6ed984798b5dc8d)

==3215629==Register values:
rax = 0x000000000000003a  rbx = 0x0000519000001980  rcx = 0x000052100012ca80  rdx = 0x0000000000000000  
rdi = 0x0000519000001af0  rsi = 0x0000000000000980  rbp = 0x00007ffd3666aa50  rsp = 0x00007ffd3666a890  
 r8 = 0x0000000000000001   r9 = 0x0000000000000001  r10 = 0x0000000000000028  r11 = 0x0000000000000000  
[...]

PR #6928, opened 2026-08-23
fix(multiheadattention): add num_heads and embed_dim validation to prevent FPE

### LINKED ISSUE #6926: [bug] MultiHeadAttention::load_param divides by zero when num_heads is 0, crashing Net::load_param on any param file that sets 1=0
MultiHeadAttention::load_param divides by zero when num_heads is 0, crashing Net::load_param on any param file that sets 1=0

[ncnn_attachment.zip](https://github.com/user-attachments/files/31326644/ncnn_attachment.zip)

## error log

```
Floating point exception (core dumped)
```

Under a debugger the faulting instruction is the integer division in

```
src/layer/multiheadattention.cpp:59
scale = pd.get(6, 1.f / sqrtf(embed_dim / num_heads));
```

## context

- ncnn at `c189d88`, still present on current `master`
- Built exactly as your own x86-64 CI builds it, `.github/workflows/linux-x64-cpu-clang.yml`:
  `-DNCNN_AVX=OFF -DNCNN_AVX2=OFF`, Release, clang
- No sanitizer and no coverage instrumentation in the build that crashes
- Linux x86-64

## how to reproduce

1. Save these four lines as `crash.param`:

```
7767517
2 2
Input                    in    0 1 in
MultiHeadAttention       mha   1 1 in out 0=64 1=0
```

2. Load it:

```cpp
#include "net.h"
{
    ncnn::Net net;
    net.load_param(argv[1]);       // crash.param
    return 0;
}
```

3. `Floating point exception`. Deterministic: 5 runs out of 5 on a plain Release build.

## more

**Root cause.** `MultiHeadAttention::load_param` reads both operands from the param file
and checks neither:

```cpp
int MultiHeadAttention::load_param(const ParamDict& pd)
{
    embed_dim = pd.get(0, 0);
    num_heads = pd.get(1, 1);
    ...
    scale = pd.get(6, 1.f / sqrtf(embed_dim / num_heads));   // line 59
```

`embed_dim` and `num_heads` are both `int`, so `embed_dim / num_heads` is an **integer**
division, and `num_heads == 0` raises SIGFPE before `sqrtf` is ever called.

**The default of 1 does not protect it.** `pd.get(1, 1)` returns 1 only when key 1 is
*absent*. A param file that sets `1=0` supplies zero, and zero is what is used. Omitting
`1=` entirely does not crash - I checked, it loads cleanly.

**Supplying `scale` does not protect it either.** This is the part worth noting: the
expression `1.f / sqrtf(embed_dim / num_heads)` is a C++ default *argument*, so it is
evaluated unconditionally at the call site, whether or not key 6 is present. I confirmed
this - `... 0=64 1=0 6=1.250000e-01` crashes identically.

**Suggested fix**, rejecting the value where it is read rather than at the division:

```cpp
num_heads = pd.get(1, 1);
if (num_heads <= 0)
{
    NCNN_LOGE("MultiHeadAttention invalid num_heads %d", num_heads);
    return -1;
}
```

A negative `num_heads` is worth rejecting in the same check; it does not trap, but it
produces a negative head dimension that nothing downstream expects.

**Same pattern, different layer.** #6911 reports the same shape of defect in
`ConvolutionDepthWise::load_param` — an integer divide by a param value that is not
checked. This one is in `MultiHeadAttention` and is not covered by that report or by
#6913, #6914 or #6915, none of which touch `multiheadattention.cpp`. It may be worth
sweeping the other `load_param` implementations for the same construct while you are in
there.

Found by fuzzing the param reader. Happy to send more detail if it is useful.

lxcxjxhx:
## Description

Fix divide-by-zero (SIGFPE) in `MultiHeadAttention::load_param` when `num_heads` or `embed_dim` is 0.

## Root Cause

```cpp
// src/layer/multiheadattention.cpp:59
scale = pd.get(6, 1.f / sqrtf(embed_dim / num_heads));
```

- `embed_dim / num_heads` is integer division
- When `num_heads = 0`, this causes SIGFPE
- The default value `1` for `num_heads` only applies when the key is absent, not when explicitly set to 0
- Even providing `scale` explicitly doesn't help because the default argument is evaluated unconditionally

## Fix

Add validation checks after reading `num_heads` and `embed_dim` from ParamDict:

```cpp
num_heads = pd.get(1, 1);
if (num_heads <= 0)
{
    NCNN_LOGE("MultiHeadAttention invalid num_heads %d", num_heads);
    return -1;
}

if (embed_dim <= 0)
{
    NCNN_LOGE("MultiHeadAttention invalid embed_dim %d", embed_dim);
    return -1;
}
```

## Testing

Verified with the reproduction case from #6926:
```
7767517
2 2
Input                    in    0 1 in
MultiHeadAttention       mha   1 1 in out 0=64 1=0
```

Before: `Floating point exception (core dumped)`
After: `MultiHeadAttention invalid num_heads 0` (returns -1)

## Related Issues

Fixes #6926
Related: #6911, #6913, #6914, #6915 (similar validation issues)

### DIFF
--- a/src/layer/multiheadattention.cpp
+++ b/src/layer/multiheadattention.cpp
@@ -52,6 +52,19 @@ int MultiHeadAttention::load_param(const ParamDict& pd)
 {
     embed_dim = pd.get(0, 0);
     num_heads = pd.get(1, 1);
+
+    if (num_heads <= 0)
+    {
+        NCNN_LOGE("MultiHeadAttention invalid num_heads %d", num_heads);
+        return -1;
+    }
+
+    if (embed_dim <= 0)
+    {
+        NCNN_LOGE("MultiHeadAttention invalid embed_dim %d", embed_dim);
+        return -1;
+    }
+
     weight_data_size = pd.get(2, 0);
     kdim = pd.get(3, embed_dim);
     vdim = pd.get(4, embed_dim);


nihui:

PR #6930, opened 2026-08-25
vulkan: apply atan2 zero workaround to packed shaders

### LINKED ISSUE #6884: atan2-vec4: any test exercising BinaryOp op_type 10/11 with pack=1 on a device using Mesa v3dv fail
## error log

```
value not match  at c:0 d:0 h:0 w:6    expect 0.000000 but got 2.356191
output blob 0 not match
test_layer_gpu failed
test_layer BinaryOp failed pack=1 fp16p=0 fp16s=0 fp16a=0 bf16p=0 bf16s=0 sgemm=1 winograd=1
test_binaryop_atan2_special failed op_type=10
```

## context

- ncnn, current `master` (commit `a4d2ea1`)
- Hardware: Raspberry Pi 4B, Broadcom VideoCore VI GPU
- Driver: Mesa `v3dv` (Mesa 25.1.8), confirmed via `vulkaninfo`: `driverID = DRIVER_ID_MESA_V3DV`
- Only reproduces with `pack=1` (the `binaryop_pack4.comp` / vectorized shader path). The scalar path (`binaryop.comp`, no packing) computes `atan(0.0, 0.0)` correctly and matches the CPU reference.

## how to reproduce
1. Run `ctest -R test_binaryop_3` (or any test exercising `BinaryOp` op_type 10/11 with `pack=1`) on a device using Mesa `v3dv`.
2. Observe the mismatch at the `(0, 0)` input element: expected `0.000000` (matching `atan2f(0.0, 0.0)`), got `2.356191` (`3π/4`).
3. Standalone confirmation, no ncnn involved: a minimal compute shader dispatching a single invocation of
   ```glsl
   vec4 v1 = vec4(0.0, 1.0, 0.0, -0.0);
   vec4 v2 = vec4(0.0, 0.0, 1.0, -0.0);
   result[0] = atan(v1, v2);
   ```
   returns `atan(0.0, 0.0) = 2.356191` and `atan(-0.0, -0.0) = -2.356191` (both wrong - CPU gives `0.0` and `-3.141593` respectively). The equivalent **scalar** call, `atan(float(0.0), float(0.0))`, correctly returns `0.0` on the same device/driver. Only the vectorized `atan(vec4, vec4)` form is affected.

## more

Three shader files contain the vec4 form and are all affected: `binaryop_pack4.comp`, `binaryop_broadcast_pack4.comp`, `binaryop_broadcast_pack1to4.comp` (byte-identical logic in all three). Each already contains a workaround for exactly this class of issue:

```glsl
#if ncnn_driverID == 13
    // llvmpipe driver (VK_DRIVER_ID_MESA_LLVMPIPE = 13) workaround for atan(0, 0), match atan2f(0, 0) == 0
    if (op_type == 10)
    {
        bvec4 zero_mask = bvec4(uvec4(equal(v1, afpvec4(0.f))) & uvec4(equal(v2, afpvec4(0.f))));
        afpvec4 v2_safe = mix(v2, afpvec4(1.f), zero_mask);
        res = atan(v1, v2_safe);
    }
    ...
#else
    if (op_type == 10) res = atan(v1, v2);
#endif
```

but it's gated to `ncnn_driverID == 13` (Mesa `llvmpipe`) specifically, so `v3dv` (a different driver ID) falls through to the unguarded call and hits the same underlying issue `llvmpipe` apparently also had.

(`binaryop.comp` and `binaryop_broadcast.comp`, the scalar - non-packed - variants, contain a similar-looking but textually different workaround using a plain ternary rather than `equal()`/`mix()`. These are *not* affected: the scalar `atan(float, float)` form computes `atan(0.0, 0.0)` correctly on this driver, confirmed separately - only the vectorized `atan(vec4, vec4)` form is broken.)

Given the workaround (substituting a safe nonzero value for the second argument only when both inputs are exactly zero) is a no-op for every other input, it doesn't need to be driver-specific at all - applying it unconditionally (or at least also for `DRIVER_ID_MESA_V3DV`) in all three affected shader files would fix this without needing to track every affected driver individually.

[002-fix-atan2-vec4-zero-zero-master.patch](https://github.com/user-attachments/files/30620378/002-fix-atan2-vec4-zero-zero-master.patch)

LEKKO2310:
## Summary

Apply the existing packed-vector `atan(0, 0)` workaround to all non-MoltenVK drivers. Mesa v3dv exhibits the same incorrect vectorized result as llvmpipe, so restricting the workaround to driver ID 13 leaves packed `BinaryOp` atan2/ratan2 broken on v3dv.

The change covers:

- `binaryop_pack4.comp`
- `binaryop_broadcast_pack4.comp`
- `binaryop_broadcast_pack1to4.comp`

## Verification

Built with Vulkan and system glslang on Ubuntu 24.04, then ran `test_binaryop_3` with Mesa llvmpipe 25.2.8:

```
100% tests passed, 0 tests failed out of 1
```

Fixes #6884.

### DIFF
--- a/src/layer/vulkan/shader/binaryop_broadcast_pack1to4.comp
+++ b/src/layer/vulkan/shader/binaryop_broadcast_pack1to4.comp
@@ -98,8 +98,6 @@ void main()
     if (op_type == 10) res = afpvec4(atan(vec4(v1), vec4(v2)));
     if (op_type == 11) res = afpvec4(atan(vec4(v2), vec4(v1)));
 #else
-#if ncnn_driverID == 13
-    // llvmpipe driver (VK_DRIVER_ID_MESA_LLVMPIPE = 13) workaround for atan(0, 0), match atan2f(0, 0) == 0
     if (op_type == 10)
     {
         bvec4 zero_mask = bvec4(uvec4(equal(v1, afpvec4(0.f))) & uvec4(equal(v2, afpvec4(0.f))));
@@ -112,10 +110,6 @@ void main()
         afpvec4 v1_safe = mix(v1, afpvec4(1.f), zero_mask);
         res = atan(v2, v1_safe);
     }
-#else
-    if (op_type == 10) res = atan(v1, v2);
-    if (op_type == 11) res = atan(v2, v1);
-#endif
 #endif
     if (op_type == 12) res = v1 - trunc(v1 / v2) * v2;
     if (op_type == 13) res = v2 - trunc(v2 / v1) * v1;
--- a/src/layer/vulkan/shader/binaryop_broadcast_pack4.comp
+++ b/src/layer/vulkan/shader/binaryop_broadcast_pack4.comp
@@ -167,8 +167,6 @@ void main()
     if (op_type == 10) res = afpvec4(atan(vec4(v1), vec4(v2)));
     if (op_type == 11) res = afpvec4(atan(vec4(v2), vec4(v1)));
 #else
-#if ncnn_driverID == 13
-    // llvmpipe driver (VK_DRIVER_ID_MESA_LLVMPIPE = 13) workaround for atan(0, 0), match atan2f(0, 0) == 0
     if (op_type == 10)
     {
         bvec4 zero_mask = bvec4(uvec4(equal(v1, afpvec4(0.f))) & uvec4(equal(v2, afpvec4(0.f))));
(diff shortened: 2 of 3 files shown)


tencent-adm:
[CLA assistant check](https://cla.opensource.tencent.com/Tencent/ncnn?pullRequest=6930) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you all sign our [Contributor License Agreement](https://cla.opensource.tencent.com/Tencent/ncnn?pullRequest=6930) before we can accept your contribution.<br/>**1** out of **2** committers have signed the CLA.<br/><br/>:white_check_mark: LEKKO2310<br/>:x: nihui<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.opensource.tencent.com/check/Tencent/ncnn?pullRequest=6930) it.</sub>

nihui (approved):

nihui:

PR #6945, opened 2026-09-02
pnnx for torch exported program

lvy010:
Make pnnx support torch dynamo exported program model conversion

```shell
pnnx model.pt2
```

Add pt2 format support to all pnnx unit tests, parse the model graph and weights in pt2 without introducing third-party libraries, and pass multi-platform compilation and testing.

---

This lets pnnx load `torch.export` exported programs (`.pt2`) directly, using the same CLI as the TorchScript path:

```shell
pnnx model.pt2
pnnx model.pt2 inputshape=[1,3,224,224]
```

The archive is parsed with pnnx's own ZIP reader and a new in-tree JSON parser, so no third-party dependency is added. The frontend builds the PNNX IR from the serialized graph and raw tensor payloads and then hands off to the existing `pass_level2`..`pass_level5` and `pass_ncnn` pipeline, so operator lowering is shared with the TorchScript path rather than reimplemented.

This is a deliberately **fail-closed** frontend: anything it cannot prove is rejected with a specific diagnostic instead of being converted approximately. The scope, the exact checks and the known gaps are documented in [`tools/pnnx/README.md#pt2-compatibility`](tools/pnnx/README.md) and [`tools/pnnx/docs/pt2-validation.md`](tools/pnnx/docs/pt2-validation.md).

```mermaid
flowchart LR
    A[model.pt / model.pt2] --> B{detect_model_format}
    B -->|data.pkl| C[load_torchscript]
    B -->|archive_format=pt2| D[exported_program_reader]
    B -->|unknown| Z[reject, no TorchScript fallback]
    D --> E[STORE-only ZIP + CRC32 + bounds]
    D --> F[pnnx_json -> typed schema 8.20]
    F --> G[schema / default / effect normalization]
    E --> H[tensor_meta -> Attribute]
    G --> I[load_exported_program -> PNNX IR]
    H --> I
    C --> I
    I --> J[pass_level2..level5 + pass_ncnn]
    J --> K[.pnnx.param/.bin + _pnnx.py]
    J --> L[.ncnn.param/.bin + _ncnn.py]
```

## Usage

```python
import torch

net = Model().eval()
x = torch.rand(1, 3, 224, 224)

program = torch.export.export(net, (x,))
torch.export.save(program, "model.pt2")
```

Generated artifacts match the TorchScript path: `model.pnnx.param/.bin`, `model_pnnx.py` with a deterministic `test_inference()`, and for the ncnn backend `model.ncnn.param/.bin` plus `model_ncnn.py`.

Unit tests reach the PT2 path through the shared helper, which converts and compares both formats against eager:

```python
from pnnx_test_utils import test_model_formats
ok = test_model_formats(net, (x,), expected, "test_F_softmax")
```

## Key changes

* **Format dispatch** — `model_format.cpp` classifies the archive by its members (`archive_format`/`archive_version` + `models/*.json` for PT2, `data.pkl`+`constants.pkl` for TorchScript, the flat `serialized_exported_program.json` layout for legacy). An unknown or corrupt archive is rejected outright; it never falls back to the TorchScript loader.
* **In-tree JSON parser** (`json.cpp`) — no new dependency. Parses the serialized graph including the quoted `inf`/`nan` floats that `torch.export` emits.
* **Archive reader** (`exported_program_reader.cpp`) — STORE-only consumed entries with CRC32 verified on every consumed record (including empty records and a stored CRC of zero), ZIP/ZIP64 physical bounds, directory counts and local/central header consistency. Encryption and duplicate entry names are rejected. `StoreZipReader` gained `get_file_compression()`/`get_file_flags()` so a record is inspected before its advertised payload is allocated. Unused compressed attachments may stay indexed; they are never read.
* **Typed schema** (`exported_program.h/.cpp`) — the graph, signature, range constraints and tensor metadata are decoded into typed structures, so a malformed field is a parse error rather than a downstream surprise. Version contract: `archive_version="0"`, ExportedProgram schema **8.20 exactly**, and `opset_version.aten` equal to the linked LibTorch `torch::jit::getMaxOperatorVersion()`. No operator upgrade/downgrade adapter is implemented.
* **Tensor materialization** — CPU strided raw storage on little-endian hosts. Sizes, nonnegative strides and offsets must fit the storage under checked arithmetic. Scalar shape `()` has one element, zero-sized dims have none. Non-zero storage offsets, zero strides and shared-storage views are accepted within bounds and materialized densely. Representable dtypes: bool, uint8, signed int8/16/32/64, float16/32/64, bfloat16 and complex32/64/128; the generated Python loader reconstructs bfloat16/complex32 from their storage bits.
* **Schema-driven normalization** (`exported_program_defaults.cpp`) — omitted ATen arguments are restored from the dispatcher schema and reordered into canonical order, so exported nodes match the existing `pass_level2` patterns instead of carrying misaligned `dim`/`keepdim`/`dtype` slots. Overloads are resolved against the real schema rather than by name mangling.
* **Local-effect normalization** — in-place operators are rewritten to their functional counterparts only through explicit dispatcher overload pairs, and only when the target is provably a whole, single-use, unaliased local tensor from a known allocator. Use counting includes nested references and graph/signature returns. External inputs, parameters, buffers, constants, views and existing aliases all fail closed. A separate, narrower exception lets a signature-authorized `TensorConstant` pass through `lift_fresh_copy` to `detach_`.
* **Guard handling** — only a concrete, proven-true `aten._assert_scalar`, or a statically verified `aten._assert_tensor_metadata`, is accepted; the importer cross-checks the operand's dtype/shape before removing the guard. Unresolved or runtime-dependent guards are rejected rather than dropped.
* **Input contracts** — the importer keeps dtype, rank, static dimensions, bare symbols, ranges and shared-symbol equality in reserved `__pt2_input_*` parameters on `pnnx.Input`, and `ir.cpp` emits explicit `ValueError` checks in the generated Python before computation. They stay active under `python -O`/`-OO`. **Raw ncnn models do not enforce these checks**, and the PT2 parameters are absent from ncnn output.
* **CLI validation** (`main.cpp`) — PT2 rejects nonempty `customop`/`moduleop` and any `device` other than `cpu` before creating outputs. `inputshape` dtype suffixes must match the exported dtype (an omitted suffix keeps it, rather than defaulting to `f32` as the other frontends do); `input`/`input2` NumPy payloads are read and validated but never executed or used for value specialization. The save stages now propagate failure instead of silently continuing.
* **Resource limits** — 512 MiB aggregate raw tensor loading (unique storage plus every dense view), 512 MiB per consumed record and per dense attribute, 256 MiB ZIP central directory, 1 Mi entries, 64 MiB cumulative entry-name bytes. These are trust bounds, not a whole-process memory cap.
* **Operator/lowering additions** — PT2-only graph forms needed by the shared pipeline: raw ATen conv1d/3d and conv_transpose1d/2d/3d patterns, `F.interpolate`, `Tensor.new_full`, `torch.full`, `ChannelShuffle`, `weight_norm`, plus `convert_half_to_float` and pooling/`F_instance_norm` fixes on the ncnn side.
* **Tests** — 7 native targets (~5.5k lines) covering the ZIP reader, JSON parser, format detection, schema, importer, local-effect proofs and the IR tensor contract, with negative cases for truncated archives, hostile `tensor_meta`, out-of-bounds offsets, aliases and escaping targets. Python suites cover real archives, dtypes, dynamic/bare symbols, generated-Python guards at `optlevel=0/1/2` crossed with Python `-O` levels, CLI contracts, and a native ncnn parity test comparing eager Torch, generated PNNX Python and `ncnn.Net` on the same input.
* **CI** (`.github/workflows/pnnx-pt2.yml`) — Ubuntu/Windows/macOS crossed with producer Torch 2.12.1 and 2.13.0, while the converter links LibTorch 2.12.1 in every job, using separate venvs so producer and linked roles stay distinct. Every job runs `pt2_frontend` plus a TorchScript/PT2 ReLU/Linear/Conv2d smoke selection; Linux also runs `pt2_backend`. Logs and the CMake cache are uploaded as artifacts.

## Verification

Local Windows x64 run of `de9ed7cc`, clean working tree. Producer Python 3.13.3 / Torch 2.12.1+cpu (ExportedProgram schema `(8, 20)`); converter linked against LibTorch 2.12.1; MSVC 14.29.30133, Ninja, Release; TorchVision 0.27.1+cpu, Transformers 4.51.3, NumPy 2.5.1, ncnn 1.0.20260526. All CTest selections used `--output-on-failure --no-tests=error`.

Registered and passing per selection:

* `-L '^pt2_frontend$' -j 4`: 14
* `-L '^pt2_backend$' -j 1`: 1
* `-L '^torchscript$' -j 8`: 390
* `-L '^pt2_operator$' -j 8`: 390

TorchAudio has no release matching Torch 2.12.1 (newest is 2.11.0), so the four `test_torchaudio_*` cases fail on import in **both** channels and account for the entire TorchScript column. Against the 386 cases this environment can actually run: TorchScript 386/386, PT2 377/386.

End-to-end chain on a Conv2d+BatchNorm2d+ReLU+AdaptiveAvgPool2d+Linear model exported with `torch.export.save`, converted with `pnnx model.pt2 inputshape=[1,3,32,32]`:

* the `.pt2` archive is STORE-only, as `torch.export.save` writes it;
* the ncnn graph lowers to 4 layers with **zero leftover ATen operators** — BatchNorm folded into `Convolution` and ReLU fused (`convrelu_0`), then `Pooling` and `InnerProduct`;
* max abs difference vs eager: **1.49e-08** for the generated PNNX Python, **7.55e-05** for native `ncnn.Net` (default `fp16=1` storage).

Two build-time probes were negative and bound what the run covers: `PNNX_TORCH_HAS_PICKLE_LOAD` failed, so the legacy tensor-dictionary path was not compiled and `test_pnnx_exported_program_legacy` is not registered; protobuf was absent, so there is no ONNX output to compare.

## Known limitations

Nine `pt2_operator` cases fail inside the converter. Each was attributed to its own diagnostic and each maps onto a limitation this contract already declares; they are recorded in [`tools/pnnx/docs/pt2-validation-results.md`](tools/pnnx/docs/pt2-validation-results.md).

* `test_mobilenet_v3_small_pt2` — `dropout_` has no registered in-place counterpart: `aten::dropout` renames `self` to `input`, which the schema-driven reorder cannot express
* `test_swin_t_pt2` — `fill_` writes a target that is not a single-use unaliased local
* `test_vit_b_32_pt2` — payloads exceed the aggregate 512 MiB budget
* `test_transformers_flaubert_attention_pt2`, `..._xlm_...` — `masked_fill_` target is not single-use local
* `test_transformers_mt5_attention_pt2`, `..._t5_...` — `add_` target is not single-use local
* `test_transformers_funnel_attention_pt2` — `aten._assert_scalar` is not a concrete boolean; runtime guards are not implemented
* `test_transformers_longformer_attention_pt2` — the producer serializes integer scalars into `aten.div.Tensor_mode`, whose schema declares a Tensor `self`

The in-place cases need general functionalization / alias-update lowering rather than a wider allocator list; the ownership proof declines them today instead of rewriting a tensor that is still shared.

Also not implemented, by design at this stage:

* DEFLATE decoding for consumed archive entries (official `torch.export.save` output is STORE and unaffected).
* Restricted pickle decoding; the legacy path uses LibTorch `torch::pickle_load` and accepts **trusted inputs only**.
* Full PyTree input/output reconstruction — the boundary is flat positional tensor inputs with empty kwargs, and a tensor/numeric scalar or nonempty flat tuple of such leaves.
* General dynamic shapes: symbolic dimensions are fixed by an explicit `inputshape`, and no runtime dynamic-shape asserts are emitted into ncnn.
* Four cases intentionally expect converter **rejection** and are pinned to exact diagnostics: `Tensor_fill`, `Tensor_slice_copy`, `Tensor_index_put` and `torch_masked_select`. Their TorchScript variants keep the normal conversion/parity expectation. A green negative test means the intended diagnostic was observed without a crash or timeout, not that the operator converted.


tencent-adm:
[CLA assistant check](https://cla.opensource.tencent.com/Tencent/ncnn?pullRequest=6945) <br/>All committers have signed the CLA.

lvy010:
Add native support for converting `torch.export.ExportedProgram`
(`.pt2`) models with pnnx, without introducing a third-party JSON
dependency.

This is a work-in-progress implementation organized as independently
reviewable commits.

#### Layer 1: JSON and archive foundation

- [x] 1.1 Add a minimal JSON parser
- [x] 1.2 Harden the JSON parser
- [x] 1.3 Detect TorchScript and PT2 archive formats
- [x] 1.4 Add archive validation tests

lvy010:
update

```mermaid
flowchart LR
    A[PT2 archive] --> B[安全 ZIP reader]
    B --> C[版本化 typed schema]
    C --> D[Dispatcher schema binder]
    C --> E[Tensor materializer]
    D --> F[mutation / alias 审核]
    E --> G[PNNX IR importer]
    F --> G
    G --> H[pass level2-5]
    H --> I[PNNX Python parity]
    H --> J[ONNX parity]
    H --> K[ncnn parity]
```

PR #6948, opened 2026-09-02
fix(spectrogram): validate n_fft/winlen in load_param

ktz03:
## Summary
- Reject `n_fft <= 0`, `winlen <= 0`, or `winlen > n_fft` before writing `window_data`, returning `-1` instead of overflowing.

Fixes #6939

## Test plan
- [ ] Load a param with `winlen > n_fft` / `n_fft = 0` and confirm `load_param` returns `-1` without ASan write OOB
- [ ] Valid spectrogram params still load and forward

Issue #6939, opened 2026-09-01
bug: heap-buffer-overflow (write) in Spectrogram::load_param (src/layer/spectrogram.cpp:61)


beilzx:
## error log
```
=================================================================
==2194917==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x5170000002c4 at pc 0x55ac08be257a bp 0x7ffe67e2f010 sp 0x7ffe67e2f000
WRITE of size 4 at 0x5170000002c4 thread T0
    #0 0x55ac08be2579 in ncnn::Spectrogram::load_param(ncnn::ParamDict const&) /src/test/results/fuzzing_ncnn/test/ncnn/src/layer/spectrogram.cpp:61
    #1 0x55abff7fe079 in ncnn::Net::load_param_bin(ncnn::DataReader const&) /src/test/results/fuzzing_ncnn/test/ncnn/src/net.cpp:1882
    #2 0x55abff8144c8 in ncnn::Net::load_param_bin(_IO_FILE*) /src/test/results/fuzzing_ncnn/test/ncnn/src/net.cpp:2222
    #3 0x55abff7b781c in main build/harness_bin.cpp:3
    #4 0x7f95e767d1c9  (/lib/x86_64-linux-gnu/libc.so.6+0x2a1c9) (BuildId: 274eec488d230825a136fa9c4d85370fed7a0a5e)
    #5 0x7f95e767d28a in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x2a28a) (BuildId: 274eec488d230825a136fa9c4d85370fed7a0a5e)
    #6 0x55abff7b75e4 in _start (/src/test/results/fuzzing_ncnn/test/ncnn/build/ncnn_parambin+0x29d5e4) (BuildId: 64b7f031d746f1deffff4534b7b2d9ba11374dfe)

0x5170000002c4 is located 0 bytes after 580-byte region [0x517000000080,0x5170000002c4)
allocated by thread T0 here:
    #0 0x7f95e7d4ef1d in posix_memalign ../../../../src/libsanitizer/asan/asan_malloc_linux.cpp:145
    #1 0x55abff881e7b in fastMalloc /src/test/results/fuzzing_ncnn/test/ncnn/src/allocator.h:62
    #2 0x55abff881e7b in ncnn::Mat::create(int, unsigned long, ncnn::Allocator*) /src/test/results/fuzzing_ncnn/test/ncnn/src/mat.cpp:331
[...]

PR #6950, opened 2026-09-02
fix(inversespectrogram): validate n_fft/winlen in load_param

ktz03:
## Summary
- Reject `n_fft <= 0`, `winlen <= 0`, or `winlen > n_fft` before writing `window_data` in `InverseSpectrogram::load_param`.
- Same failure class as Spectrogram ASan #6939 (fixed in #6948).

## Test plan
- [ ] Load param with invalid `winlen`/`n_fft` and confirm `-1` without heap overflow
- [ ] Valid InverseSpectrogram params still load and forward


nihui:

PR #6967, opened 2026-09-07
Fix SIGFPE with missing or incomplete CPU topology

### LINKED ISSUE #6961: SIGFPE on linux where ncnn cannot read the CPU topology
ref https://github.com/nnstreamer/nnstreamer/commit/012f873d9e897a351a336c449836ae4dff1e212b

ref https://github.com/nnstreamer/nnstreamer/pull/4379

nihui:
Fixes #6961.

## Problem

CPU topology information may be unavailable or only partially accessible in restricted environments such as chroots.

The Linux physical CPU count already falls back to the logical CPU count when no sibling information can be resolved. However, partially readable topology can still produce an incomplete physical CPU count.

Shared cache detection has a separate division-by-zero path: if cache information is readable but the physical cores sharing that cache cannot be resolved, `shared_physical_cpu_count` remains zero and is used as a divisor. This can raise SIGFPE during CPU information initialization.

## Changes

- Fall back to the logical CPU count when topology information is incomplete, instead of using a partial physical CPU count. Also apply the positive-count fallback after platform-specific detection.
- Return cache detection failure when the required sibling information cannot be resolved, and guard against a zero shared physical CPU count. This lets the caller use the existing cache-size fallback.
- Add basic sanity checks for CPU counts and the final L2 cache size. These check the values after fallback; they do not require successful hardware topology detection or an exact cache size.

The fix stays in CPU information detection rather than adding workarounds to individual convolution or GEMM implementations.

## Validation

Standalone `src/cpu.cpp` validation using mocked procfs/sysfs data:

- Five baseline scenarios reproduced SIGFPE before the fix: missing, empty, or invalid topology; an empty shared CPU map; and a shared CPU map containing only CPUs outside the detected range.
- All 11 scenarios passed after the fix, covering the above cases, partially missing topology, missing sysfs, missing or empty cpuinfo, and normal SMT/non-SMT topology. Validation included builds with and without OpenMP, plus UBSan.

The fault-injection harness was used separately and is not included in this PR. The added `test_cpu.cpp` checks are basic API sanity checks, not fault-injection tests.

Full ncnn builds, model inference tests, and testing in an actual Tizen/GBS chroot have not been performed.


tencent-adm:
[CLA assistant check](https://cla.opensource.tencent.com/Tencent/ncnn?pullRequest=6967) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.opensource.tencent.com/Tencent/ncnn?pullRequest=6967) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.opensource.tencent.com/check/Tencent/ncnn?pullRequest=6967) it.</sub>

PR #6974, opened 2026-09-08
fix(rnn): reject non-positive num_output/hidden_size in load_param

### LINKED ISSUE #6973: bug: RNN/LSTM/GRU load_model divides by zero when num_output/hidden_size is 0
## error log

```
Floating point exception (core dumped)
```

## context

- ncnn current `master`
- Same validation class as #6911 / #6970 (`load_param` accepts 0, later integer division SIGFPEs)

`RNN::load_model` / `GRU::load_model` do:

```cpp
int size = weight_data_size / num_directions / num_output;
```

`LSTM::load_model` does:

```cpp
int size = weight_data_size / num_directions / hidden_size / 4;
```

`load_param` reads these from the param file with defaults but never rejects `<= 0`. An explicit `0=0` (or LSTM `3=0`) therefore crashes during `load_model` instead of returning an error.

## how to reproduce

```
7767517
2 2
Input    in   0 1 in
RNN      rnn  1 1 in out 0=0 1=0
```

```cpp
#include "net.h"
{
    ncnn::Net net;
    return net.load_param(argv[1]); // may pass; crash on load_model / combined load
}
```

With a matching empty/zero-weight bin, or any path that reaches `load_model`, the `/ num_output` raises SIGFPE. Same for `GRU` and for `LSTM` with `3=0`.

## expected

`load_param` should reject non-positive `num_output` (and LSTM `hidden_size`) with `-100` before `load_model` divides.

## more

PR incoming covering RNN / LSTM / GRU.

ktz03:
## Summary
- Reject `num_output <= 0` in `RNN` and `GRU` `load_param` (their `load_model` divides by `num_output`).
- Reject `num_output <= 0 || hidden_size <= 0` in `LSTM` `load_param` (divides by `hidden_size`).

Fixes #6973

## Test plan
- [ ] `RNN` / `GRU` param with `0=0` → `load_param` returns `-100`, no SIGFPE
- [ ] `LSTM` param with `0=0` or `3=0` → `-100`, no SIGFPE
- [ ] Valid positive sizes still load

### DIFF
--- a/src/layer/gru.cpp
+++ b/src/layer/gru.cpp
@@ -26,6 +26,12 @@ int GRU::load_param(const ParamDict& pd)
 #endif
     }
 
+    if (num_output <= 0)
+    {
+        // reject invalid num_output (load_model divides by it)
+        return -100;
+    }
+
     return 0;
 }
 
--- a/src/layer/lstm.cpp
+++ b/src/layer/lstm.cpp
@@ -27,6 +27,12 @@ int LSTM::load_param(const ParamDict& pd)
 #endif
     }
 
+    if (num_output <= 0 || hidden_size <= 0)
+    {
+        // reject invalid sizes (load_model divides by hidden_size)
+        return -100;
+    }
+
     return 0;
 }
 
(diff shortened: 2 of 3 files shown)


nihui:

PR #6982, opened 2026-09-09
riscv: fix fp16 tanh_ps saturating at 1.044 instead of 1.0

songxiaoyuwyze:
rvv_mathfun_fp16s.h reused the single precision rational approximation coefficients from the fp32 tanh_ps. In fp16 alpha_9 alpha_11 alpha_13 underflow to zero and alpha_7 becomes subnormal, and those are exactly the terms that pull the ratio back to 1 for large z = x**2. Without them the approximation saturates at 1.044 for |x| >= 9 and is already 2 per mille high at |x| ~ 4. alpha_1 = beta_0 ~ 4.9e-3 also pushes the numerator into the fp16 subnormal range for small x, costing 48 ulp below 0.1.

Refit for half precision, still pure fp16 arithmetic:
- saturation point 9.0 -> 4.5, tanh(4.5) = 0.99975 already rounds to 1.0 in fp16, so |x| >= 4.5 returns -/+1 through a merge instead of relying on the rational form to saturate
- P3/Q3 refitted on [0, 4.5], all coefficients exactly representable in fp16 and in the normal range, alpha_1 = beta_0 = 1 removes the subnormal numerator for small x
- clamp the result to <= 1 to prevent overshoot inside the interval
- rename the macros to c_tanh_f16_* so they no longer collide with the fp32 macros of the same name in rvv_mathfun.h; tanh_riscv_zfh.cpp, mish_riscv_zfh.cpp, unaryop_riscv_zfh.cpp and riscv_activation.h all include both headers, so the fp16 values were silently redefining the fp32 ones
- 2 fma for the numerator and 2 for the denominator, 5 fewer than before

Max error in fp16 ulp, per operation fp16 rounding simulation:

  |x|        <0.1  0.1~1  1~2  2~3  3~3.5  3.5~4.5  >=4.5
  before     48.4    2.6  2.7  4.3    5.1      8.9   90.0
  after       2.3    2.9  2.5  3.3    3.4      3.3    0.5

This is the riscv counterpart of #6979, which fixed the same bug in neon_mathfun_fp16s.h. It affects TanH, Mish and UnaryOp on the fp16 arithmetic path; GELU on riscv computes tanh in fp32 and was not affected.

### DIFF
--- a/src/layer/riscv/rvv_mathfun_fp16s.h
+++ b/src/layer/riscv/rvv_mathfun_fp16s.h
@@ -272,69 +272,67 @@ _RVV_FLOAT16_COS_OP(2, 8)
 _RVV_FLOAT16_COS_OP(4, 4)
 _RVV_FLOAT16_COS_OP(8, 2)
 
-#define c_tanh_tiny 1e-4f
-#define c_tanh_hi   9.0f
+#define c_tanh_f16_tiny 1e-4f
+// tanh(4.5) = 0.99975 already rounds to 1.0 in half precision,
+// so anything outside [-4.5, 4.5] is -/+1.0f
+#define c_tanh_f16_hi 4.5f
+// tanh(x) ~= x * P(x^2) / Q(x^2), rational approximation fitted on [0, 4.5] for half precision.
+// The single precision coefficients used by the fp32 tanh_ps cannot be represented in fp16:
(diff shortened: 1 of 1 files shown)


tencent-adm:
[CLA assistant check](https://cla.opensource.tencent.com/Tencent/ncnn?pullRequest=6982) <br/>All committers have signed the CLA.

nihui:

PR #6984, opened 2026-09-09
remove woa linux ci

mingshi2333:
Remove the `woa-linux` job that builds and runs Windows ARM64 tests under Wine on Linux. Remove the `NCNN_ARM82=OFF` override from the native Windows ARM64 test job so it uses ncnn's default ARM feature detection and retains runtime coverage of supported ARM82 paths.

Validation: [native Windows ARM64 run](https://github.com/mingshi2333/ncnn/actions/runs/34409391334) using the same build and CTest commands passed all 179 tests. The validation also confirmed `NCNN_ARM82=ON` and Windows FP16/DOTPROD support. YAML parsing and `git diff --check` pass.

### DIFF
--- a/.github/workflows/windows-arm.yml
+++ b/.github/workflows/windows-arm.yml
@@ -88,7 +88,7 @@ jobs:
     - name: build
       run: |
         mkdir build; cd build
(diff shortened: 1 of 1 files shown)


nihui:

PR #6989, opened 2026-09-10
harden net loading

nihui:
Validate graph counts, blob indexes, shape hints, feature masks and layer connection counts in both text and binary parameter loaders. Reject cyclic binary graphs with iterative traversal, handle missing producers during extraction, and honor registered layer destroyers on parameter-load failures and CPU replacement paths.

Layer/blob counts and per-layer connection counts are limited to 1,000,000. These limits bound the corresponding allocation sizes, not total parsing time or aggregate memory use. Repeated blob references remain supported. For custom layer registration by integer index, the registry slot after removing `LayerType::CustomBit` must be below 1,000,000.

Compatibility and reload behavior:

- Binary custom layer type indexes must carry `LayerType::CustomBit`. Untagged unknown types no longer fall back to custom registry slots; `ncnn2mem` already adds the tag.
- A registered built-in override whose creator is null or returns null causes parameter loading to fail. Built-in fallback is used only when no override is registered, preventing a registered destroyer from receiving an object created by the built-in factory.
- Except for `Input`, layers with `one_blob_only = true` must have exactly one bottom and one top. Zero-output layers remain supported with `one_blob_only = false`.
- Entering parameter parsing clears the previous graph. A parsing failure leaves the Net empty; failure to open a parameter file returns before parsing and preserves the previous graph.

Layer destruction and pipeline destruction remain separate: parameter-load failure cleanup releases layers and graph metadata before pipelines have been created, while `Net::clear()` destroys pipelines before releasing layers.

Validation: native `test_net`, `test_paramdict` and SqueezeNet inference passed. Net tests cover malformed and truncated input, shape layouts, reloads, custom layer lifetime, CPU replacement and binary graph cycles. Deep-graph tests validate iterative cycle detection during loading; they do not establish a safe depth for recursive inference.


tencent-adm:
[CLA assistant check](https://cla.opensource.tencent.com/Tencent/ncnn?pullRequest=6989) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.opensource.tencent.com/Tencent/ncnn?pullRequest=6989) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.opensource.tencent.com/check/Tencent/ncnn?pullRequest=6989) it.</sub>

PR #7000, opened 2026-09-13
cmake: fix CMAKE_EXE_LINKER_FLAGS typo in emscripten flags

### LINKED ISSUE #6993: Typo in CMakeLists.txt (CMAKE_EXECUTBLE_LINKER_FLAGS)
https://github.com/Tencent/ncnn/blob/c0abf4830d8f8637efa5d34fce3985ba3d8642bd/CMakeLists.txt#L806-L818

Hi, I just wanted to point out a silent bug in the build system. I believe it should be `CMAKE_EXE_LINKER_FLAGS` instead of `CMAKE_EXECUTBLE_LINKER_FLAGS` in two places here.

Somebody already pointed that out over 3 years ago in #4379 but it does not appear like the PR will be merged.

yuefdev:
Fixes #6993.

The Emscripten block in `CMakeLists.txt` appends its `-s` options to `CMAKE_EXECUTBLE_LINKER_FLAGS`. CMake never reads that misspelled variable, so those two lines had no effect. This renames them to `CMAKE_EXE_LINKER_FLAGS`, which is what they were meant to set.

Wasm builds should behave the same as before. CMake already passes `CMAKE_C_FLAGS` / `CMAKE_CXX_FLAGS` on the link line, so the same `-s` options reached `emcc` through that path. With this fix the options are also set in the variable that is actually meant for linking, so they no longer depend on that side effect. Repeating an `-s` option with the same value is harmless for `emcc`.

I did not build this locally with Emscripten, so I'm relying on the CI wasm jobs for that.

AI disclosure: I made this change with help from Claude Code and reviewed it.


tencent-adm:
[CLA assistant check](https://cla.opensource.tencent.com/Tencent/ncnn?pullRequest=7000) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you all sign our [Contributor License Agreement](https://cla.opensource.tencent.com/Tencent/ncnn?pullRequest=7000) before we can accept your contribution.<br/>**1** out of **2** committers have signed the CLA.<br/><br/>:white_check_mark: yuefdev<br/>:x: nihui<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.opensource.tencent.com/check/Tencent/ncnn?pullRequest=7000) it.</sub>

nihui:

PR #7007, opened 2026-09-16
cmake: verify xtheadvector target macro in compiler probe

ihb2032:
Some toolchains accept `-march=*xtheadvector` and compile the generic vector intrinsic probe without actually selecting the XTheadVector target, so CMake may enable the XTheadVector code paths incorrectly.

Keep the existing `test()` and `test2()` intrinsic checks and append `test3()` referencing `__riscv_xtheadvector`. This verifies the same target macro that the RISC-V layer sources use while remaining compatible with the Xuantie C906/C910 toolchain.

This replaces the invalid `th.vmv1r.v` check from the previous revision. The [official XTheadVector specification](https://github.com/XUANTIE-RV/thead-extension-spec/blob/d16533b5a9cd790c369d849ddf9a48159a5a235c/xtheadvector.adoc#L42-L48) explicitly states that XTheadVector has no whole-register move instructions. Xuantie GCC generates `__riscv_xtheadvector` from the selected ISA extension list.

Replaces #6681, which could not be reopened after its branch was recreated.


nihui:
This patch caused the c906/c910 toolchain to lose xtheadvector support, see https://github.com/Tencent/ncnn/actions/runs/35078259485/job/104769962108?pr=7007

For old behavior, see https://github.com/Tencent/ncnn/actions/runs/35062331143/job/104685091287

ihb2032:
Thanks, fixed in 53cb7756.

The previous `th.vmv1r.v` probe was invalid: the official XTheadVector specification explicitly says there are no whole-register move instructions, so the C906/C910 toolchain correctly rejected it.

`test3()` now references `__riscv_xtheadvector` instead. Xuantie GCC V3.4.0 generates this ISA macro when `xtheadvector` is selected, and ncnn's RISC-V layer sources use the same macro for their XTheadVector-specific paths. This keeps the existing intrinsic checks and avoids depending on assembler syntax differences between toolchain generations.

ihb2032:
The updated workflows are currently waiting for maintainer approval after the force-push. Could you please approve the new CI run? I will verify that the C906/C910 `NCNN_COMPILER_SUPPORT_RISCV_XTHEADVECTOR` probe is successful on commit 53cb7756.

nihui (approved):

nihui:

PR #7008, opened 2026-09-16
fix: bounds checking for heap overflow in CopyTo, Padding, Tile, ModelBin

0x4r35:
## Summary

Add input validation for attacker-controlled parameters parsed from model files (`.param` / `.bin`) to prevent heap buffer overflows during inference.

- **CopyTo** (`src/layer/copyto.cpp`): Validate `woffset`/`hoffset`/`doffset`/`coffset` against destination Mat dimensions before copy. Prevents arbitrary out-of-bounds heap write via crafted offset parameters.

- **Padding** (`src/layer/padding.cpp`): Reject negative padding values in `forward()`. Negative `left` causes `memcpy(outptr + left, ...)` to write before the allocated buffer (heap underflow). Special auto-pad values `-233` (SAME_UPPER) and `-234` (SAME_LOWER) are preserved.

- **Tile** (`src/layer/tile.cpp`): Use 64-bit multiplication to compute output dimensions (`w * repeat_w`, etc.) and check for `INT_MAX` overflow before allocating. Prevents small allocation + large write via integer overflow.

- **ModelBin** (`src/modelbin.cpp`): Use 64-bit multiplication in `load(w, h, type)`, `load(w, h, c, type)`, and `load(w, h, d, c, type)` to detect `w * h * ...` overflow before passing to `load(int, type)`. Prevents corrupted Mat dimensions from undersized allocation.

All four issues are triggerable by loading a crafted `.param` file and confirmed with AddressSanitizer.

## Test plan

- [ ] Verify existing unit tests pass (`ctest`)
- [ ] Confirm normal models with valid parameters still load and infer correctly
- [ ] Confirm models using `-233`/`-234` auto-padding still work
- [ ] ASAN build with crafted `.param` files no longer triggers heap overflow


tencent-adm:
[CLA assistant check](https://cla.opensource.tencent.com/Tencent/ncnn?pullRequest=7008) <br/>All committers have signed the CLA.

0x4r35:
recheck

0x4r35:
@codex review

0x4r35:
recheck

PR #7015, opened 2026-09-20
x86: reduce Col2im memory access for non-overlapping deconvolution

Hosi121:
The x86 SGEMM deconvolution path first fills the output with bias, then reads and updates each value in Col2im. When dilation is 1 and stride equals the kernel size, each output has only one source. This change writes the source plus bias once and fills only the output padding.

The path supports rectangular kernels and packs of 1, 4, and 8. When there are fewer output channel groups than threads, it also splits input rows between workers. Each worker writes separate output rows. The existing crop and activation steps remain in place. Other shapes and pack 16 use the existing path.

There is no new work tensor. The GEMM work, Col2im buffer size, and per-value addition order stay the same.

### Performance

Complete FP32 Deconvolution calls, including SGEMM, measured against master `1857775c98921d71e1ec2b34dc7d292c72529423`. Intel Core Ultra 7 255H, WSL2, GCC 13.3, Release, AVX512 disabled. Each result is the median of three fresh processes; each process uses nine samples of 15 calls after eight warmup calls. The run order changes between rounds. Input and output packing for the probe are outside the timed region. These are operator results, not full-model results.

| Kernel/stride; input HxW; input/output channels | Threads | Master ms | PR ms | Time reduction |
|---|---:|---:|---:|---:|
| 2x2; 184x184; 16/16 | 1 | 1.697 | 1.198 | 29.4% |
| 2x2; 184x184; 16/16 | 4 | 0.699 | 0.360 | 48.5% |
| 2x2; 368x368; 16/16 | 4 | 5.899 | 2.938 | 50.2% |
| 4x4; 96x96; 16/16 | 4 | 0.705 | 0.294 | 58.2% |
| 2x3; 128x128; 16/16 | 4 | 0.410 | 0.222 | 45.9% |
| 3x3; 128x128; 8/4 | 4 | 0.237 | 0.083 | 65.1% |
| 2x2; 184x184; 16/16; right/bottom padding 1 | 4 | 0.667 | 0.327 | 51.0% |

The overlap control retains the old loop. Its measured time increased by 1.7% with one thread and 2.2% with four threads. All 26 conditions, raw samples, exact shapes, build options, and the comparison code are in the [separate benchmark branch](https://github.com/Hosi121/ncnn/tree/364251e19c363a54d6dbfc7a6086585f147fcd35/benchmarks/col2im). No ARM, Vulkan, AVX512, or full-model speed result is claimed.

### Validation

- Added 32 cases to `test_deconvolution` for rectangular kernels, output padding, packing, and one or four threads. The complete test passed.
- Compared complete forward output bytes with master for 236 configurations and 1,433,578 values. All matched. These include bias, activation, crop padding, odd sizes, and fallback cases.
- The exact Col2im loops passed ASan/UBSan for 300 configurations and 16,375,872 values, including guard bytes and special float values. This uses a small Mat substitute, not a full sanitizer build of ncnn.
- The changed-line format check and `git diff --check` passed.

The [reproduction instructions](https://github.com/Hosi121/ncnn/blob/364251e19c363a54d6dbfc7a6086585f147fcd35/benchmarks/col2im/README.md) use the normal ncnn shared library and public layer API. The benchmark files are outside this PR diff.

This change, its tests, and this description were prepared with help from an AI assistant.


tencent-adm:
[CLA assistant check](https://cla.opensource.tencent.com/Tencent/ncnn?pullRequest=7015) <br/>All committers have signed the CLA.

PR #7018, opened 2026-09-21
pnnx: write blob shape hints into the ncnn param

Edwardssss:
### Description

A model converted by `pnnx` carries no shape information at all. Shapes reach a loaded model through exactly one place: param index `30`, spelled `-23330` in the text param, which `Net::load_shape_hints()` reads into `Layer::bottom_shapes` / `top_shapes` (`src/net.cpp`). That parameter is written by `ncnnoptimize` today (`tools/modelwriter.h:890`), but `ncnnoptimize` cannot repair a pnnx model either: `ModelWriter::shape_inference()` seeds its dummy inputs from the `Input` layer shape parameters (`0=w`, `1=h`, `2=c`, `11=d`) and gives up when they are missing (`tools/modelwriter.h:396`, `Input layer ... without shape info, shape_inference skipped`), and `pnnx` writes none of those parameters. Measured on a pnnx export: the `ncnnoptimize` output has no hint and prints that message; adding `0=32 1=32 2=3` to the `Input` layer makes the inference run and it writes back exactly the records this PR emits.

Three consumers show what the missing shapes cost:

* 26 Vulkan layer files choose the element packing variants they create with `shape.dims == 0` (`cast_vulkan.cpp:65`: `if (shape.dims == 0 || shape.elempack == 1)`, likewise `concat_vulkan`, `flatten_vulkan`, `pooling_vulkan`, ...). Without a shape they create all of them instead of the one the tensor will use.
* Layers that specialize on the shape stay generic: `softmax_vulkan.cpp:28` takes `top_shapes[0]` as its workspace shape, and `convolution_vulkan.cpp:92-113` builds `shape_bordered` from it and passes `w` / `h` / `cstep` as shader specialization constants (`:501-503`); an empty shape leaves them at 0.
* `ncnn_blob_get_bottom_shape()` / `ncnn_blob_get_top_shape()` report `dims = 0` for every blob, so a tool that merely loads the model cannot read any shape.

This PR emits the hints from the traced shapes pnnx already has, in the encoding that `tools/modelwriter.h` writes and `src/net.cpp::load_shape_hints()` reads:

* one record per output, `step = 5` for every record of a layer, so 3-dim and 4-dim outputs can coexist
* the batch axis (`__batch_index`, `233` = no batch axis) is dropped, which is the same rule the generated test uses to feed `ncnn.Mat(numpy, batch_index=N)`, so the hint describes the tensor the runtime is actually given
* the remaining axes map to `Mat(w,h,d,c)` in reverse order: torch `[1,3,32,32]` becomes `dims=3, w=32, h=32, c=3`
* a layer whose output shapes are not all known gets no hint at all, so partially traced models behave exactly as before
* the previously undocumented format is now described in `docs/developer-guide/param-and-model-file-structure.md`, and `tools/pnnx/tests/test_pnnx_shape_hint.py` checks the emitted records

### Validation

Environment: Debian (WSL2), 8 threads, ncnn built with `-DNCNN_VULKAN=OFF` and `NCNN_VALIDATION=ON` (so `load_shape_hints()` validates every record), pnnx built against torch 2.13, tolerance `1e-4`.

Generated param for `Conv2d(3,8,3x3) + ReLU + AdaptiveAvgPool2d(1) + Linear(8,4)`, input `[1,3,32,32]`:

```
Input              in0        0 1 in0      -23330=5,3,32,32,1,3
Convolution        convrelu_0 1 1 in0 1    -23330=5,3,32,32,1,8
Pooling            gap_0      1 1 1 2      -23330=5,1,8,1,1,1
InnerProduct       linear_2   1 1 2 out0   -23330=5,1,4,1,1,1
```

The change only appends a parameter, so what needs checking is the code that reads and uses the param:

| Check | Result |
|:---|:---|
| `test_c_api` — public C API (Mat, net load, extractor) | pass |
| `test_net` — param loading, covers `load_shape_hints` for the text and the binary param | pass |
| `test_paramdict` — param encoding, including the `-23300` old array form and malformed records | pass |
| `test_mat_batch` — the `Mat::n` batch semantics that the batch-axis rule relies on | pass |
| `test_cpu` — cpu set / OpenMP / powersave runtime checks | pass |
| a param id the runtime does not know (`-23305=4,1,2,3,4` added next to the hints) | loads, output unchanged |
| `ncnnoptimize` reading and rewriting a param that carries the hints | hints preserved |

The param with the hints, and the same param with every hint removed, give the same result (3.494e-05 `max_abs_diff`, 0.013 ms `best of 200`), so the parameter does not change what the CPU path computes; an `fp16=0` export of the same model gives `5.601e-05`, i.e. the residual comes from the two runtimes, not from the hints. The x86 convolution only reads the shapes to choose which winograd variant to pre-transform, and only when `prefer_winograd` is active (`num_input > 8 || num_output > 8`); `forward()` re-derives them from the real `Mat`.

The new `test_pnnx_shape_hint` checks the emitted records; it exits 1 when an expected record is perturbed, so its assertion is live. A hand written decoder check prints `layer->top_shapes` for 1, 2, 3 and 4-dim records, for a record with `dims=0` and for a layer with two outputs of different rank, and the 4-dim case is also visible through `ncnn_blob_get_top_shape()` (`dims=4 w=32 h=32 c=16`) — the C API keeps its existing `dims, w, h, c` contract.

### Notes

* Vulkan was not measured here, so this is a source reading. Hints do not trigger the intentional CPU fallback: every `support_vulkan = false` site is driven by layer params, never by `bottom_shapes` / `top_shapes`. They do change which packing the layers expect: `net.cpp` derives it from the hint (`c % 4 == 0 ? 4 : 1` for 3/4-dim, `h % 4 == 0 ? 4 : 1` for 2-dim, `w % 4 == 0 ? 4 : 1` for 1-dim) and converts the incoming blob to it (`vkdev->convert_packing`), so a hint that matches the tensor keeps the pipelines consistent and replaces the zeros with the real shape. The hazard is a **stale** hint, i.e. running a model at a shape other than the one it was traced with; a shape-hinted `ncnnoptimize` model already has that contract, and dropping the parameter restores the unknown-shape behaviour.
* `ncnn_blob_get_bottom_shape()` / `ncnn_blob_get_top_shape()` only expose `dims, w, h, c`, so `d` is not reachable from the C API.


Edwardssss:
This PR adopt a conservative scheme: the shape parameter already exists (`-23330`, param index `30`), `ncnnoptimize` already writes it and `Net` already reads it, including the 4-dim form (`step = 5` with the `d` slot) since #6553. What was missing is the converter, so `pnnx` now writes the same records and a freshly exported model no longer starts with empty `top_shapes`.

If that is not what you expect, the alternatives I can give are shown below:

* **this PR.** The hint stays the only shape carrier, and every tool that knows the shapes writes it (`ncnnoptimize`, and now `pnnx`). No new parameter, and runtimes that do not know index `30` keep ignoring it.
* **write the input shape on the `Input` layer.** Emit `0=w` / `1=h` / `2=c` / `11=d` in the param, so a model declares its input shape without a hint and without a first run. `pnnx` currently writes none of these parameters, and `Input` already accepts `d` as param `11`.
* **expose `d` through the C API.** `ncnn_blob_get_shape()` / `ncnn_blob_get_bottom_shape()` / `ncnn_blob_get_top_shape()` still return only `dims, w, h, c`, so a 4-dim shape is not readable from C at all. 

Which direction do you prefer? And I wonder if other better ideas exists, if so,I am glad to change in this PR.

